PR #3, opened 2018-12-03
Prevent docs from being uploaded from PR builds

drasmuss:
**Motivation and context:**

We check if `TRAVIS_BRANCH==master` in order to decide whether to update `gh-pages`.  However, there's a bit of a gotcha with Travis where PR builds will have `TRAVIS_BRANCH=master` (rather than the name of the PR branch).  This generally doesn't affect us, because we don't enable PR builds.  But in case these bones are used in a repo that does enable PR builds, this makes sure everything works smoothly.

**How has this been tested?**

Implemented these changes in NengoDL (which does have PR builds), and they work as advertised (e.g., note that https://travis-ci.org/nengo/nengo-dl/jobs/462924853 does not upload the docs).

**Types of changes:**

**Checklist:**


drasmuss:
This was superseded by #5

PR #7, opened 2019-04-05
Add more documentation

drasmuss:
**Motivation and context:**

Fills out the README, and adds some more rendered documentation (including a notebook demonstrating the different configuration options).

I also added a `generate-bones all` option (since that seemed like a good idea as I was writing up the usage instructions).

Based on #5 

**How has this been tested?**

Built the docs and examined them locally.  Note that TravisCI reports the doc build as failing, but if you check the log you can see that it's just the link checker that's failing (because these new docs contain links to pages that don't exist yet), not the actual doc build.

**Types of changes:**

**Checklist:**

**Still to do:**


docs/conf.py
```diff
@@ -12,6 +12,7 @@
     "nengo_sphinx_theme",
     "numpydoc",
     "nbsphinx",
+    "sphinx_click.ext",
```
arvoelke: Ah, we each found this same extension. Great. :)

docs/cli.rst
```diff
@@ -0,0 +1,26 @@
+Command line interface
+======================
+
+In general, the only command that downstream repos will ever need to run is
+
+.. code-block:: bash
+
+    generate-bones all
```
arvoelke: Is it important to note that the defaults won't put the scripts in the right place? Should the defaults be changed? Also related to what happens when `check-bones` fails due to version mismatch: it simply says to run `generate-bones`.

drasmuss: You're right, should put something here about which directory to be in when calling `generate-bones`, and should update that message in `check-bones` now that the `all` option exists.

arvoelke: Actually have you tested `all` on the other repos? I just tried it on `nengo-loihi` by cd'ing to `.ci` and running:

```
generate-bones --conf-file=../.nengobones.yml --template-dir="" all
```

It puts `.codecov.yml` and `.travis.yml` in the wrong place (inside `.ci` instead of one directory back). Also note that the empty string is needed in order to add the cwd to the templating paths. Ideally I think all that a user should need to do is run:

```
generate-bones all
```

from the root directory, and then it will use appropriate defaults for `--template-dir` and `--output-dir` depending on if it's `codecov-yml`, `travis-yml`, or the `ci-scripts`.

docs/cli.rst
```diff
@@ -0,0 +1,26 @@
...
+
+This updates all the templated files that need to be manually updated.  Note
+that there are other files that ``nengo-bones`` generates, but they are generated
+automatically by ``nengo-bones`` as needed, and will not need to be manually
+updated or committed to the downstream repo.
```
arvoelke: Oh I missed the significance of this note. I read it and thought "yeah yeah" but didn't realize that this meant that "all" is not actually "all". The "all" setting is just the two root yml files for now. Maybe "all" could be renamed to "local", and then make it clear it should be run from the root directory to get the right defaults?

docs/cli.rst
arvoelke: Never mind the above ^ this was confusion on my part. See the below note.

docs/cli.rst
arvoelke: Pushed a commit that improves this part of the documentation. Also changed "all" to the default command, as per offline discussion.

arvoelke:
Added a couple commits to:
 - change `all` to the default command
 - improve some of the documentation / error messages
 - raise a better error message if missing `.nengobones.yml` (also unit-tested)

LGTM. After #5 is merged I'll rebase this and fix up the history.

arvoelke:
By the way I checked out the docs and they look good. I just had to fixup a couple commits to update the logo and update the names used in the example notebook to the new `ci_scripts` / `travis_yml` / `codecov_yml` formats. LGTM.

nengo_bones/templates/.travis.yml.template
```diff
@@ -74,7 +74,7 @@ before_install:
   - export -f _travis_terminate_unix
   - export -f _travis_terminate_windows
   # install/run nengo-bones
-  - pip install git+https://github.com/nengo/nengo-bones.git@template-ci
+  - pip install git+https://github.com/nengo/nengo-bones.git@docs
```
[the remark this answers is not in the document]
tbekolay: Will this be updated to `pip install nengo-bones` or will be continue to test with `master`? It makes sense to test with the release version on other repos, but not necessarily for testing in this repo.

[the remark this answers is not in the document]
drasmuss: Ah yeah good call. I'd say we make this part templated `pip install {{ bones_install }}`, with a default of `nengo-bones`, and then for this repo we can set that to master instead.  

Actually it'd be even better if we could have it automatically set to the current branch (as it's easy to forget to update this when you start working on a new branch, and then be very confused about why none of your changes are working. not that that happened to me 👀 ). Which I think we could do by just setting `bones_install` to `git+https://github.com/nengo/nengo-bones.git@${TRAVIS_PULL_REQUEST_BRANCH:-$TRAVIS_BRANCH}`.

.nengobones.yml
```diff
@@ -24,6 +24,8 @@ travis_yml:
   deploy_dists:
     - sdist
     - bdist_wheel
+  bones_install: >-
+    git+https://github.com/nengo/nengo-bones.git@$(git symbolic-ref --short HEAD)
```
drasmuss: This goes past the 80 character limit with or without the linebreak, so I'd probably just put it all on one line to avoid the extra continuation character stuff.  Either way though 🤷‍♂️

.travis.yml
```diff
@@ -80,7 +80,7 @@ before_install:
     _travis_terminate_unix
     _travis_terminate_windows
   # install/run nengo-bones
-  - pip install git+https://github.com/nengo/nengo-bones.git@$(git rev-parse --abbrev-ref HEAD)
+  - pip install git+https://github.com/nengo/nengo-bones.git@$(git symbolic-ref --short HEAD)
```
drasmuss: Just a note for merge, this should be fixing up "Make nengo-bones version configurable", not "Add more documentation", right?

docs/introduction.rst
```diff
@@ -1 +1,6 @@
+************
+Introduction
+************
+
 .. include:: ../README.rst
+   :start-line: 16
```
drasmuss: It feels like there's no way we're going to remember to update this (or at least no way I'm going to remember), and then our introduction page is going to be broken until someone happens to notice.  That possibility outweighs the benefit of skipping the badges I'd say.  Maybe we could add a big note in `README.rst` saying "don't forget to go update `introduction.rst`"?  Although even then it seems like it'd be pretty easy not to notice that comment if you're touching a different part of the readme.

tbekolay: The benefit is more skipping the Nengo Bones header than the badges but yeah let me try the other method of doing this to see if it can be done.

PR #16, opened 2019-04-22
Static checks only delete .py files they created

hunse:
In checking the notebooks, we turn them into .py files. We now
only delete the .py files associated with notebooks, and not
other .py files that might be in the `docs/examples` directory.

**Motivation and context:**

In nengo-loihi, we have other .py files in `docs/examples` that we should not delete.

**How has this been tested?**

Tried it out locally.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo_bones/templates/static.sh.template
+++ b/nengo_bones/templates/static.sh.template
@@ -24,10 +24,18 @@ shopt -s globstar
     #   s/# $/#/g replaces lines with just '# ' with '#'
     #   /get_ipython()/d removes lines containing 'get_ipython()'
     #   $ d removes a trailing newline
-    sed -i -e 's/# $/#/g' -e '/get_ipython()/d' -e '$ d' -- docs/examples/**/*.py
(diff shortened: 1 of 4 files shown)


nengo_bones/templates/static.sh.template
```diff
@@ -25,7 +25,9 @@ shopt -s globstar
     #   $ d removes a trailing newline
     sed -i -e 's/# $/#/g' -e '/get_ipython()/d' -e '$ d' -- docs/examples/**/*.py
     exe pylint docs --rcfile=setup.cfg --disable=missing-docstring,trailing-whitespace,wrong-import-position,unnecessary-semicolon
-    rm docs/examples/**/*.py
+    for notebook_file in docs/examples/**/*.ipynb; do
+        rm "${notebook_file%.ipynb}.py";
```
tbekolay: I think the semicolon at the end here isn't necessary?

hunse: Removed

PR #17, opened 2019-04-23
Template other files

hunse:
**Motivation and context:**

We want templating for other files.

**How has this been tested?**

Currently just hand-testing on some other repos. PRs for other repos will reference this issue and appear below as they're created.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] contributing
- [x] manifest
- [x] setup.cfg
- [x] docs/conf.py
- [x] setup.py
- [x] Add documentation for all this
- [x] Add tests for all this
- [x] Do we want nengo-bones header on rst files?
- [x] Update this repository's .nengobones.yml and regenerate files


arvoelke:
Did we also want to template out `pull_request_template.md`?

hunse:
So one reason I made some config options like "author" appear both in the highest-level `config` and sub-configs like `license_config`, rather than passing both `config` and `license_config` to the license template like 2a46d2b918b29b3501be9d67e9dbb38c6facbf5a does, is so that license can overwrite the author if it wants. That way we could have e.g. "Applied Brain Research" as the author in setup.py, and "Applied Brain Research, Inc." as the author in LICENSE.rst.

drasmuss:
Yeah I was thinking about that, but my vote would be to keep the configuration options as simple as we can.  So unless we find a really compelling reason where we'd need to override top-level config options, just set those things in one place.  I don't think we've needed that flexibility yet, so this way we just have a single source of truth for, e.g., the `author` value.

hunse:
I did use it for "description" in Nengo, because we have a longer description in setup.py than in the docs. But we can probably figure out a shorter description for Nengo that will work everywhere.

drasmuss:
Yeah I'm just looking through all the other repos now to see if any of my fixups broke anything (I'm sure they did), will take a look.

hunse:
One question I had when writing this is what do we want to do with the `__copyright__` variable currently in `__init__.py` in some of our repos. It'd be nice if this didn't have to be updated manually, but it seems like a pain to add templating to `__init__.py` just for this. Can we just get rid of these? The main place I saw them used was in `docs/conf.py`, and we got rid of the use there.

nengo_bones/templates/docs.sh.template
```diff
@@ -5,7 +5,7 @@
 {% block install %}
 {{ super() }}
     exe pip install "sphinx>=1.8.0" "jupyter>=1.0.0" "nbsphinx>=0.2.13" "nengo-sphinx-theme>=0.7.0" "numpydoc>=0.6.0"
-    exe pip install -e ".[docs]"
+    exe pip install -e ".[doc]"
```
tbekolay: Why did we change these to `doc` and `test` instead of `docs` and `tests`? I think up this points we've been using the plural, so switching to singular would mean changing documentation if we recommend using these somewhere, and will be confusing if someone was previously using one of these and then has to switch.

drasmuss: I changed the variables in `setup.py` to `doc_req`, `test_req` etc., because I always found it confusing keeping track of which were plural and which weren't (e.g. `install_requires` vs `docs_require` vs `optional_requires`).  So I standardized on the singular, since that seemed less grammatically awkward (e.g. `installs_req`). And I changed from `requires` to `req` to make it feel less like a normal word that we'd expect to be pluralized one way or the other.  And then since those were all changing to singular, it seemed like it might be confusing to have the install names be pluralized.  I didn't expect anyone to be using these other than developers, so I didn't think there'd be too much user impact from swapping.  But we could also leave it as plural, and not worry about the mismatch between `[docs]` and `doc_req` (which people might not even notice or be confused by, unless they're inspecting the `setup.py` code).

tbekolay: The reason it's done with singular/plurals in those spots is to be consistent with `setuptools` itself, as the `setup` function accepts `install_requires`, `tests_require` and `extras_require` keyword arguments.

But we're kind of free to do what we want for our own naming things as long as we pass things to the correct keyword argument, so I'm fine with `_req` instead of `_require(s)`. For `doc`/`test` vs `docs`/`tests`, however, I would rather be consistent with other places in our source tree; e.g., we have a top-level `docs` directory, not a `doc` directory, and within `nengo_bones` and other places we have `tests` directories, not `test`, so I would find it confusing to switch to singular here.

nengo_bones/config.py
```diff
@@ -56,22 +127,27 @@ def validate_config(config):
...
+                    break
+                else:
+                    raise KeyError("Config file must define %s" % entry)
 
     if "ci_scripts" in config:
         for ci_config in config["ci_scripts"]:
             validate_ci_config(ci_config)
 
+    if "license_rst" in config:
+        check_list(config["license_rst"], "append")
```
tbekolay: Is this key actually valid? I can't see it referenced in LICENSE.rst.template, did it get replaced by extending the template?

drasmuss: Ah yeah that's been removed, forgot to take that part out.

drasmuss: (sidenote, there are a bunch of other keys that we could be applying this check to, but I haven't gone through and done that systematically)

.nengobones.yml
```diff
@@ -124,12 +124,10 @@ setup_cfg:
       - tests/ignoreme
     markers:
       test-marker:
-        Not a real
-        marker
+        Not a real marker
     nengo_test_unsupported:
-      nengo/tests/*:
-        All nengo tests are
-        unsupported
```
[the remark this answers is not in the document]
tbekolay: Ah gotcha, will revert.

nengo_bones/scripts/generate_bones.py
```diff
@@ -161,39 +161,34 @@ def travis_yml(ctx):
 @click.pass_context
 def codecov_yml(ctx):
     """Generate codecov config file."""
-
```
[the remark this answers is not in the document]
tbekolay: Normally I do too, but looked weird for a one-line function ... but also not that passionate about it, will revert

PR #18, opened 2019-04-24
Add pr-number script

tbekolay:
**Motivation and context:**
When we make a changelog entry, we add a link to the pull request for additional context. If the pull request doesn't exist yet, we need to figure out what number the PR will end up being. This script gets that number and formats it as a URL that can be used in changelogs.

**How has this been tested?**
This script doesn't have robust error checking or unit tests yet; it can be considered WIP, but I wanted to make it so I didn't forget to make it later.

Here it is in action, run from the `nengo-bones` repo top-level directory.

```bash
~/C/nengo-bones (nengo) pr-number
Asking GitHub for information about nengo/nengo-bones...
If you open a PR now it will be assigned #18
Use this link for the changelog entry:
https://github.com/nengo/nengo-bones/pull/18
~/C/nengo-bones (nengo) pr-number nengo/nengo
Asking GitHub for information about nengo/nengo...
If you open a PR now it will be assigned #1526
Use this link for the changelog entry:
https://github.com/nengo/nengo/pull/1526
```

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Tests
- [x] Add to Sphinx docs


nengo_bones/tests/test_pr_number.py
```diff
@@ -0,0 +1,56 @@
...
+from click.testing import CliRunner
+import pytest
+
+from nengo_bones.scripts import pr_number
+from nengo_bones.tests.utils import assert_exit, make_has_line, write_file
+
+
+def test_pr_number_repo(monkeypatch):
+    # monkeypatch so we don't rely on Github connection
+    monkeypatch.setattr(pr_number, "get_issue_count", lambda repo: 9)
```
tbekolay: I think it's better to monkeypatch `requests.get` so that we limit the amount of stuff that we're mocking out. We can also then test that we're requesting from the correct URL and that we correctly handle the JSON payload that we get back. It makes it more obvious what we have to change if Github changes how they do things, but still doesn't require Github to be up and running for us to do our tests.

tbekolay: We also wouldn't have to do a `pragma: no-cover`

PR #32, opened 2019-05-15
Clear notebook

hunse:
**Motivation and context:**

This is a [commonly used script](https://github.com/ctn-waterloo/devscripts/blob/master/clearoutputs.py) from the devscripts repository. Bringing it here to make it more easily accessible.

Currently based on #17, only last commit is unique.

**How has this been tested?**

TODO

None of the core code has changed from the old script. Just added a Click wrapper.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] add tests
- [x] document in Sphinx docs


hunse:
One thing I don't like about using `black` on the code cells is that it removes semicolons from the ends of lines, which can be useful for suppressing output in cells.

tbekolay:
> One thing I don't like about using `black` on the code cells is that it removes semicolons from the ends of lines, which can be useful for suppressing output in cells.

There are other ways to suppress output in cells, like calling `plt.show()` as the last line in the cell.

hunse:
It might be a good idea to add a command-line option for applying black (could be true by default), since notebooks in some repositories might not adhere to this yet.

nengo_bones/templates/static.sh.template
```diff
@@ -8,11 +8,11 @@ shopt -s globstar
 
 {% block install %}
 {{ super() }}
-    exe pip install "jupyter>=1.0.0" "pylint>=1.9.2,<2.5" "codespell>=1.12.0" "gitlint>=0.1.2" "collective.checkdocs>=0.2" "flake8>=3.7.7"
+    exe pip install "jupyter>=1.0.0" "pylint>=2.5.1" "codespell>=1.12.0" "gitlint>=0.1.2" "collective.checkdocs>=0.2" "flake8>=3.7.7"
 {% endblock %}
 
 {% block script %}
-    exe pylint {{ pkg_name }} --rcfile=setup.cfg --jobs=0
```
hunse, line 15: I've got `pylint==2.5.3` on my machine, and I still need to use the `--rcfile` when I run things locally (I tried upgrading to 2.6.0, and same thing). If I run with `--verbose`, it says "No config file found", so somehow it's not detecting the pylint section in `setup.cfg`. (I'm testing with nengo, running from the project root folder.)

hunse: Ahh, if I run `bones-generate` and get the `[pylint.*]` sections added (as added below), it uses it properly without `--rcfile`. I guess we don't have any projects that use the CI scripts without using nengo-bones to generate `setup.cfg`, so this should be fine. At the same time, I don't think there's any harm to being explicit and keeping the `--rcfile` in.

drasmuss: Removing the explicit `--rcfile` allows the command to be run more flexibly (allows pylint to search for the `setup.cfg` file, rather than requiring it to be in the same directory). It was also kind of a hack the way we were doing it before (when `setup.cfg` wasn't supported by `pylint`, you could trick it into using one anyway through the `rcfile` option). Now that they officially support `setup.cfg`, we should always be using the correct configuration style (which has the `[pylint.*]` sections).

PR #33, opened 2019-05-16
Add more visible failure message to shell scripts

drasmuss:
**Motivation and context:**

It can be hard to pick out failures when reading through a long log on e.g. TravisCI.  This makes them more obvious with an explicit failure message and a bright red colour.

**How has this been tested?**

Added a test commit so you can see what a failure looks like here https://travis-ci.org/nengo/nengo-bones/jobs/533351613#L783

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo_bones/templates/base_script.sh.template
+++ b/nengo_bones/templates/base_script.sh.template
@@ -15,7 +15,7 @@ exe() {
     for i in "${!args[@]}"; do
       [ -n "${args[$i]}" ] || unset "args[$i]"
     done
-    "${args[@]}" || STATUS=1;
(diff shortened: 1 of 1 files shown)


hunse:
Personally, I prefer the light red. It's bolder and easier to read on my (black background) terminal. The TravisCI terminal background is a little lighter, so it's less of a difference, but I still think it's a bit better. See both here: https://travis-ci.org/nengo/nengo-bones/jobs/533491922#L829

I've made that change here, so if people are okay I'll go ahead and merge. Everything else LGTM. I don't think this is big enough to warrant a changelog entry.

drasmuss:
I'm good with either one!

PR #36, opened 2019-05-24
Add codespell_skip option

hunse:
**Motivation and context:**

So we can add directories (e.g. `_vendor`) for codespell to skip.

**How has this been tested?**

Tests here, and tested on `nengo` so that I can add `_vendor` to the skip list there.

**Types of changes:**

**Checklist:**


hunse:
Also, I should mention that I did test the case where no `codespell_skip` option is provided, and it works as expected. Calling `set` on Jinja's null type in a filter just returns the empty set. I can add a unit test for that if people want.

hunse:
Ok

Issue #40, opened 2019-05-31
Cannot find nengo-bones on Travis CI anymore


jgosmann:
See for example this build: https://travis-ci.org/nengo/nengo-spa/jobs/539885439
Even though nengo-bones 0.2.1 gets installed, the generate-bones command is not found. It is not clear to my why at the moment. It was working previously (with an install from master), but stopped working after I made changes to unrelated Jupyter notebooks. Then I changed the config to install from pip without any change.

jgosmann:
I probably figured it out: I was still using an older version locally. It might be a good idea to ensure that the scripts that are generated locally and checked in use the same nengo-bones version on the CI system that was used to generate them. Those scripts should not break just because a newer version was released.

jgosmann:
Or maybe I did not figure it out. [This job](https://travis-ci.org/nengo/nengo-spa/jobs/539889921) fails with the same error and seems to call the correct command now (bones-generate).

tbekolay:
The failure is because the TravisCI job is using Python 2.7 so it still installs an older version of nengo-bones, a version before we set `python_requires>=3.5`.

jgosmann:
So is there no way to still support Python 2.7 and 3.4 when using nengo-bones?

drasmuss:
That's correct.  We're dropping support for Python<3.5 across the nengo ecosystem, but we haven't gotten around to upgrading all the repos yet.  It'll probably happen at the same time as we roll out nengo-bones to more repos.

drasmuss:
> Those scripts should not break just because a newer version was released.

This is an intentional part of nengo-bones' design.  The idea is to enforce a common standard across the nengo ecosystem, so we want an explicit error whenever a repo is not up to date.  That being said, the error should be the informative one provided by `bones-check`, not the syntax error you were getting (due to the name changing from `generate-bones` to `bones-generate`). But that's due to the fact that nengo-bones is still in the alpha phase where there aren't really any guarantees of API stability, that will settle down once we due a 1.0 release.

jgosmann:
> We're dropping support for Python<3.5 across the nengo ecosystem

Not sure if I want to do this for Nengo SPA already. Nengo 3 hasn't been released yet and Nengo SPA is still supporting Nengo 2.7/2.8 which also supports Python 2.7.

I think ideally the Python version to run the test infrastructure should be independent of the Python version under which the tests are run, but I suppose that might be non-trivial.

> > Those scripts should not break just because a newer version was released.
> 
[...]

Issue #41, opened 2019-06-10
bones-generate succeeds, but build fails due to non-generated files


jgosmann:
Currently, running `bones-generate` gives me the following output
```
No config entry detected for docs_conf_py, skipping
No config entry detected for setup_cfg, skipping
No config entry detected for setup_py, skipping
No config entry detected for contributing_rst, skipping
No config entry detected for contributors_rst, skipping
No config entry detected for license_rst, skipping
No config entry detected for manifest_in, skipping
```
and a 0 exit code which looks like the missing config entries are just warnings that can be safely ignored. However, on Travis CI Nengo bones then complains that these files are either not existing or not generated with Nengo bones and fails the build ([see this build for example](https://travis-ci.org/nengo/nengo-spa/jobs/543616379)).

To provide a better developer experience according to the fail fast principle, `generate-bones` should indicate that these are hard errors (non-zero exit code and message that those config section are required) or the Travis CI build should continue anyways.

drasmuss:
This is just due to a bug, having non-versioned files shouldn't be an error. Will fix!

drasmuss:
Fixed in #45

PR #45, opened 2019-07-16
Check for file content rather than version number

drasmuss:
**Motivation and context:**

Previously ``bones-check`` relied on version numbers to determine if a file was up-to-date. This is a bit fragile, because for development versions it is possible to have files with the same version number but different content. And it meant that all files needed to be updated on a new nengo-bones release, even if that release didn't alter the content of the files.

In this approach we check the actual content of the generated files (by generating a temporary version of the file and comparing it to the file stored in the repository).

One substantive change with this is that previously bones-generated files could be manually edited post-creation, whereas now ``bones-check`` would raise an error for that.  That might be considered a feature? But I think it's something we want to discourage. It also meant that it was possible to accidentally edit those files, without getting an error in ``bones-check``.

This also fixes https://github.com/nengo/nengo-bones/issues/40 and https://github.com/nengo/nengo-bones/issues/41

The docs build is currently failing due to a change in `nengo-sphinx-theme`, but that is being worked on in a separate PR.

**How has this been tested?**

Updated ``test_check_bones``.

**Types of changes:**

**Checklist:**


tbekolay:
Does the generated content include the `# Version:` line? Should we just remove all of those from the templates, or explicitly skip checking those lines?

drasmuss:
Yeah it does check that line as well at the moment. I think I'd vote for just removing that line from our templates (since the content of the file is no longer really attached to the version number).

tbekolay:
Sounds good to me. If there's ever any question about what version of the templates were used, we can surely figure it out using the git history.

nengo_bones/scripts/generate_bones.py
```diff
@@ -131,128 +90,87 @@ def ci_scripts(ctx):
...
-        st = os.stat(path)
-        os.chmod(path, st.st_mode | stat.S_IEXEC)
+        output_file = params.pop("output_name", script_name)
+        BonesTemplate(script_name + ".sh").render_to_file(
+            ctx.obj["output_dir"],
+            output_name=output_file + ".sh",
+            version=__version__,
+            pkg_name=config["pkg_name"],
+            repo_name=config["repo_name"],
+            **params,
```
drasmuss: This logic could go in `flatten_config` (renamed to e.g. `get_render_data`) along with the other template-specific render argument logic.

nengo_bones/scripts/generate_bones.py
```diff
@@ -131,128 +90,87 @@ def ci_scripts(ctx):
...
-                version=nengo_bones.__version__, **params))
-
-        # Mark CI script as executable
-        st = os.stat(path)
-        os.chmod(path, st.st_mode | stat.S_IEXEC)
+        output_file = params.pop("output_name", script_name)
+        BonesTemplate(script_name + ".sh").render_to_file(
+            ctx.obj["output_dir"],
+            output_name=output_file + ".sh",
+            version=__version__,
```
drasmuss: Note: remove this after the commit removing version numbers from templates

nengo_bones/scripts/generate_bones.py
```diff
@@ -1,49 +1,28 @@
...
-        f.write(template.render(
-            version=nengo_bones.__version__,
-            **kwargs,
-            # pass in the top-level config options as well
-            # TODO: separate "top-level" config into its own section?
-            **ctx.obj["config"],
-            **ctx.obj["config"][cfg_key],
-        ))
+    kwargs = flatten_config(ctx.obj["config"], template.section)
+    kwargs["version"] = __version__
```
drasmuss: This is already set in `render_to_file`

nengo_bones/templates.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # We mark all `.sh` files as executable
+        if output_name.endswith(".sh"):
+            st = os.stat(output_path)
+            os.chmod(output_path, st.st_mode | stat.S_IEXEC)
+
+
+def _load_env():
+    bones_toplevel = os.path.normpath(
+        os.path.join(os.path.dirname(__file__))
```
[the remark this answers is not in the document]
tbekolay: Yes! It used to have a `".."`, which I removed, but forgot to remove the `os.path.join`. Fixed.

nengo_bones/templates.py
```diff
@@ -0,0 +1,121 @@
...
+        loader=jinja2.ChoiceLoader([override_loader, builtin_loader]),
+        trim_blocks=True,
+        lstrip_blocks=True,
+        keep_trailing_newline=True,
+    )
+    jinja_env.filters["rstrip"] = lambda s, chars: s.rstrip(chars)
+    return jinja_env
+
+
+env = _load_env()
```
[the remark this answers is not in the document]
tbekolay: It's not really either, it's just for convenience and shorter code. We could replace it all the uses of it with calls to `load_env` instead, that would be fine by me, just seemed like we have fewer worries about people using it in weird multithreaded situations so might as well do it with the fewest characters. But we can replace with calls to `load_env` if you prefer that.

nengo_bones/config.py
```diff
@@ -218,3 +218,75 @@ def construct_mapping(loader, node):
     fill_defaults(config)
 
     return config
+
+
+def flatten_config(config, section):
```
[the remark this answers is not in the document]
tbekolay: Makes sense! I wish I had a better name than `get_render_data`, but I can't think of it now so I'll use that.

.nengobones.yml
```diff
@@ -149,6 +149,7 @@ setup_cfg:
 docs_conf_py:
   exclude_patterns:
     - tests/test-example.ipynb
+    - examples/LICENSE.rst
```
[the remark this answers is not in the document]
tbekolay: You're right, that's what happened. I'll remove this.

nengo_bones/scripts/generate_bones.py
tbekolay: CI scripts are a bit more complicated since their output filename doesn't correspond to a section name ... I think more refactoring would have to happen to get CI scripts working the same way, so for now I'd vote to leave this as is.

nengo_bones/scripts/generate_bones.py
tbekolay: Not sure where you think it should be removed ... currently it gets removed in the same commit all the other version stuff is done ?

nengo_bones/scripts/generate_bones.py
tbekolay: I don't think it is set in `render_to_file`? It's a moot point anyhow since it gets removed, but I'm pretty sure it only gets set here.

drasmuss: You're right, I'm not sure what I was looking at 🤷‍♂

nengo_bones/templates.py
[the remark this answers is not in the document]
drasmuss: Added a fixup to do this, I've just been burned too many times by global state, I'm paranoid :wink:. I used the click context so that the environment only needs to be set up once per `bones-check/generate` call.

drasmuss:
Fixups look good to me

Issue #48, opened 2019-07-24
Apply black to notebooks


tbekolay:
Currently, our Black support is only for Python files and does not extend to Jupyter notebooks, which also contain Python code. There are several tools out there for applying `black` formatting to Jupyter notebooks, most of which are only accessible through the Jupyter notebook or Jupyterlab environment. A notable exception is [black-nb](https://github.com/tomcatling/black-nb), which looks to do almost exactly what we want -- I was able to reformat a notebook and check it through the command line.

However, it is a pretty small project and so some features are missing. Notably, it is not compatible with the most recent version of black, and it cannot currently be easily used as a pre-commit hook. I've [made issues on that project](https://github.com/tomcatling/black-nb/issues) to track those issues. If they're solved quickly then we should adopt `black-nb`; if not, we should think about either helping maintain the project, fork it, or some similar solution. Note that the project is only around 500 lines of Python code, so helping out should not be a big burden for us.

PR #49, opened 2019-07-24
Adopt Black and check conformance in downstream repos

tbekolay:
**Motivation and context:**
[Black](https://github.com/psf/black) automatically handles a lot of Python style decisions that we've spent time talking about in the past, freeing up our time for more important matters.

The only thing that I wasn't able to do in this PR is auto-format notebooks, though doing so is very close with [black-nb](https://github.com/tomcatling/black-nb). There are only a few minor issues with it, and rather than deal with them now, I chose to defer that to later. See #48 for details.

Any other PR that touches Python code will have to be reformatted with Black. It shouldn't be very difficult though, as most of the changes in this PR are either straightforward or can be ignored and re-applied by running `black` again.

**How has this been tested?**
Ran the static tests locally with and without black formatted code to ensure that checks are raised correctly.

**Types of changes:**

**Checklist:**


docs/style.rst
```diff
@@ -11,6 +11,25 @@ and follow it when working on Nengo projects.
...
 
+We use ``black`` for automated formatting.
+We recommend that you set up your editor to run ``black``
+when you save a Python file;
+see `Black editor integration
+<https://github.com/psf/black#editor-integration>`__
+for details on how to do that.
+
+Whether you set up ``black`` with your editor or not,
+we require that you run ``black`` through a pre-commit hook.
```
drasmuss: Do we need to require this? It seems like if someone prefers to do it through editor configurations (or even to run black manually), it doesn't make much difference to us. Since we're double checking that everything was run correctly in the static checks, in the worst case the only pain will fall on the contributor (who will have to re-run black and re-commit).

docs/examples/configuration.ipynb
```diff
@@ -1611,6 +1611,64 @@
...
+    "\n",
+    "The `.pre-commit-config.yaml` and `pyproject.toml` files\n",
+    "make it easy to install `black`\n",
+    "as a pre-commit hook for your repository\n",
+    "and run it on the correct files.\n",
+    "\n",
+    "The only configuration available for these files is an\n",
+    "`exclude` option that accepts a list of patterns to exclude.\n",
+    "Each entry is a regular expression.\n",
+    "The list should be the same for both files.\n",
```
drasmuss: We should add a check in `validate_config` that the lists match.

.nengobones.yml
```diff
@@ -150,9 +150,13 @@ docs_conf_py:
...
+    suppress_warnings: '["image.nonlocal_uri"]'
   extensions:
     - sphinx_click.ext
   intersphinx_mapping:
     click: https://click.palletsprojects.com/en/7.x
   analytics_id: UA-41658423-2
+
+pre_commit_config_yaml: {}
+
+pyproject_toml: {}
```
drasmuss: Should test the `exclude` parameter in both of these.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+repos:
+-   repo: https://github.com/psf/black
+    rev: stable
+    hooks:
+    - id: black
+      exclude: |
+          (?x)(
+                docs/conf.py
+              | setup.py
+          )
```
drasmuss: Would it be possible to do this through a `nengo-bones` command, rather than in every repo?  Something like `bones-pre-commit`, which would add our `nengo-bones` defined pre-commit hooks.  Advantages being that it only needs to be done once and then will work in all projects in that environment, reduces the config file clutter in downstream repos, and removes the need for the redundant `exclude` lists in `pre_commit_config_yaml` and `pyproject_toml`. But I'm not sure if it's possible to do it like that 🤷‍♂.

nengo_bones/templates/pyproject.toml.template
```diff
@@ -0,0 +1,16 @@
+# Automatically generated by nengo-bones, do not edit this file directly
+
+[build-system]
+requires = ["setuptools", "wheel"]
```
drasmuss: What is this part for?

nengo_bones/templates/pyproject.toml.template
```diff
@@ -0,0 +1,16 @@
...
+
+[tool.black]
+target-version = ['py35']
```
drasmuss: Should we include py36/py37 here as well? I thinkkk all our nengo-bones-controlled projects support all 3?

nengo_bones/templates/.pre-commit-config.yaml.template
```diff
@@ -0,0 +1,15 @@
...
+    hooks:
+    - id: black
+      exclude: |
+          (?x)(
+                docs/conf.py
+              | setup.py
+              {% for pattern in exclude %}
+              | {{ pattern }}
+              {% endfor %}
+          )
```
drasmuss: Shouldn't black pull the exclude settings from `pyproject.toml` when it is run?

nengo_bones/templates/pyproject.toml.template
```diff
@@ -0,0 +1,16 @@
...
+exclude = '''
+(
+      docs/conf.py
+    | setup.py
```
drasmuss: Random thought, would it be possible to run our templates through black during the rendering process (so our templated files would always be black (and probably pylint/flake8) compliant)?

.nengobones.yml
tbekolay: It's used in the configuration notebook. Not sure it's worth adding unnecessary files for testing these very simple options, but I can.

docs/style.rst
tbekolay: We don't need to, but I do think it's good practice because the pre-commit hook is very fast while pushing and waiting for a TravisCI build is not very fast. The pre-commit hook only runs the check on the files that are modified rather than the whole repository. I think if people get used to running pre-commit hooks (which are super simple once you do it once) then we can add more stuff in there and save ourselves time long-term.

That's my opinion anyhow, this is something I'd probably let the consensus decide.

.pre-commit-config.yaml
tbekolay: Technically (at the moment) it isn't possible to not duplicate the `exclude` lists (see https://github.com/psf/black/issues/438). We could reduce the duplication by having a more general list that both files read, or be even more general with some kind of `black` config section that doesn't correspond to a file -- we could do similar things with `flake8` and whatnot as well, but those would be bigger changes to how Bones works so I think would require more thought.

I can't really see a way to implement something like `bones-pre-commit` that doesn't also involve having these files and running `pre-commit install` because `pre-commit install` makes changes to the current repository's metadata (specifically the `.git/hooks/` directory). If we wanted to manage the pre-commit stuff ourselves, we'd still have to have some kind of configuration and run `bones-pre-commit` in every repo. `pre-commit` is also a pretty big / mature / stable project in its own right (https://github.com/pre-commit/pre-commit) so reimplementing it doesn't feel like a good use of time.

nengo_bones/templates/pyproject.toml.template
tbekolay: `pyproject.toml` is a general file for doing Python project configuration; it can completely replace `setup.cfg` and `setup.py` using tools like [Flit](https://flit.readthedocs.io/en/latest/) or [Poetry](https://poetry.eustace.io/). This little piece of metadata tells `pip` that, despite having a `pyproject.toml` file, we are still using the normal `setuptools` and `wheel` tools to distribute our packages. It's part of the [PEP-518 spec](https://www.python.org/dev/peps/pep-0518/#build-system-table).

It is optional, though, we can remove it if we don't want to specify it. It's then undefined what tools reading `pyproject.toml` do if they want to infer the build system. There is, unfortunately, no other way to configure Black.

nengo_bones/templates/pyproject.toml.template
tbekolay: Yeah sounds good. I wasn't too sure how this settings works, I was mostly trying to make sure Black doesn't use a bunch of Python 3.6+ only features, which it will do if it can't automatically infer that you're using Python 3.5.

nengo_bones/templates/.pre-commit-config.yaml.template
tbekolay: It does not when `pre-commit` runs Black because `pre-commit` only runs on the files that were modified, and passes them in directly to the `black` CLI. The exclude patterns are only run when recursing directories; if you pass in an excluded file directly it will still run Black on it. https://github.com/psf/black/issues/438 has more details.

nengo_bones/templates/pyproject.toml.template
tbekolay: Yeah, I don't see why not! in the worst case we can open a subprocess, but I bet there's a programmatic way to do it by import `black`. It would mean another dependency, but I'm fine with that. I'll see how easy it is and push a commit with it that we can skip if we don't like it in the end.

tbekolay: Pushed that commit, I think it worked out pretty cleanly.

tbekolay:
Pushed some commits to address Dan's review comments.

docs/style.rst
krisfrenette: Not to jump in on this, but I think that the pre-commit should be a required step. If they don't want to set it up in their editor, then I think that should be up to them, although I don't see why they wouldn't want to.

hunse: I agree with Dan that I don't think we need to require this, though we should encourage it perhaps. I see the `black` check just like our other static checks; it certainly helps to somehow do them yourself before you commit, but it's not a requirement. I guess the downside to this is that somebody could have a commit that fails `black` that is fixed by a later commit, and ideally we'd have every commit passing `black` (and all the other static checks for that matter).

When it comes down to it, though, I'm not sure if it really matters what language we use here. There's no way to enforce that somebody does it, and if we're strongly encouraging people to use it, I'm fine to just say it's "required". Within our team I think we should require it. People who are infrequent contributors to Nengo might not even be reading this document, and I don't think using the word "required" is going to deter anybody from contributing to Nengo because of the hassle of setting up pre-commit hooks.

drasmuss: Pushed a fixup to slightly change the wording. Still indicating that it is an expected part of our workflow, just not quite as "aggressive" as making it a requirement.  Feel free to revert if you want to stick with the "required" language, I'm fine either way!

.nengobones.yml
drasmuss: Added this in a fixup, see below.

.nengobones.yml
```diff
@@ -103,6 +103,7 @@ setup_py:
   # note: docs/tests_req are intentionally empty to test that the correct
   # requirements are built into the ci scripts
   install_req:
+    - black>=19.3b0
```
drasmuss: Turns out that black only supports Python>=3.6, so this would mean nengo-bones will no longer support 3.5. Do we want to make that change? Or just drop this black autoformatting?

drasmuss: In either case we'll need to make more changes if we want to support 3.5 (probably using conda to create a separate Python>=3.6 environment in which to run the black commands during the static checks). But it also means in general that no one working in a 3.5 environment will be able to conform to our style guide.

nengo_bones/tests/test_config.py
```diff
@@ -62,6 +62,21 @@ def test_validate_config():
...
+        config.validate_config(init_cfg)
+    init_cfg["pre_commit_config_yaml"] = {}
+
+    # error when Black exclude lists don't match
+    init_cfg["pre_commit_config_yaml"]["exclude"] = ["file2.py"]
+    init_cfg["pyproject_toml"]["exclude"] = ["file1.py"]
+    with pytest.raises(ValueError, match="must have the same 'exclude' list"):
+        config.validate_config(init_cfg)
+    del init_cfg["pre_commit_config_yaml"]["exclude"]
+    del init_cfg["pyproject_toml"]["exclude"]
```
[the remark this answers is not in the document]
tbekolay: Just to put the `init_cfg` back into a non-failing state in case we add more things to this test.

nengo_bones/templates.py
```diff
@@ -117,7 +118,23 @@ def render(self, **data):
...
+        if self.output_file.endswith(".py"):
+            black_mode = FileMode(
+                target_versions={
+                    TargetVersion.PY35,
+                    TargetVersion.PY36,
+                    TargetVersion.PY37,
+                },
+                line_length=88,
+                is_pyi=False,
+                string_normalization=True,
```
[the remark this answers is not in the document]
tbekolay: I thought these were required; if not then we should definitely only specify what we need to specify. The class is defined [here](https://github.com/psf/black/blob/master/black.py#L183)

.nengobones.yml
tbekolay: Yeah good point ... I think I'm fine with requiring 3.6 in Nengo Bones but having downstream repos continue to support 3.5 (even though devs would need access to Python 3.6).

nengo_bones/templates.py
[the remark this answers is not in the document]
drasmuss: Added a fixup to remove them.

nengo_bones/templates/.pre-commit-config.yaml.template
```diff
@@ -9,7 +9,7 @@ repos:
       exclude: |
           (?x)(
               {% for pattern in exclude %}
-              {% if not loop.first %}| {% endif %}{{ pattern }}
+              {{ "" }}{% if not loop.first %}| {% endif %}{{ pattern }}
```
tbekolay: What is the `{{ "" }}` for?

drasmuss: For some reason the `if` block doesn't preserve indentation, so the `{{ "" }}` is just kind of a noop (which will be indented), and then the if block is anchored against that.

.nengobones.yml
drasmuss: Added a fixup so that things will run on 3.5, but anything related to black is skipped (which should be fine as long as downstream repos have at least some builds running >=3.6)

CHANGES.rst
```diff
@@ -19,14 +19,22 @@ Release History
...
 **Added**
 
 - Added style guide and release instructions to documentation. (`#44`_)
+- Added templates for ``.pre-commit-config.yaml`` and ``pyproject.toml``
+  so downstream repositories can easily adopt Black. (`#49`_)
+
+**Changed**
+
+- We now check that Python source files are autoformatted with Black
+  in the ``static.sh`` script. (`#49`_)
```
tbekolay: Should add a changelog entry for auto-formatting all Python templates

nengo_bones/templates/.pre-commit-config.yaml.template
tbekolay: :+1:

tbekolay:
Fixups all look good to me! Only thing that'd be nice before merge is to add that changelog entry, which I'll leave to you @drasmuss.

PR #53, opened 2019-08-02
Add pydocstyle to static checks

tbekolay:
**Motivation and context:**
When verifying https://github.com/nengo/nengo/pull/1557, it seemed easier to just enable the checks that were being implemented there than do manual verification that all instances of the issues were fixed. And if we feel like they're worth making a Nengo PR for, we might as well add those checks to Nengo Bones so that we don't add more docstrings that would fail those checks.

That said, this PR likely needs some discussion; see the todo section at the bottom.

**How has this been tested?**
Applied to Nengo Bones.

**Types of changes:**

**Checklist:**

- [ ] I have read the **CONTRIBUTING.rst** document.
- [ ] I have run the test suite locally and all tests passed.

**Still to do:**

- [ ] Come to consensus on what checks we want to add / ignore (most notably `D401`, see the squash commit in this PR)
- [ ] Add documentation
- [ ] Add changelog entry

### DIFF
--- a/nengo_bones/templates/setup.cfg.template
+++ b/nengo_bones/templates/setup.cfg.template
@@ -115,10 +115,28 @@ plt_filename_drop =
 {% endif %}
 {% endblock %}
 
+{% block pydocstyle %}
+[pydocstyle]
+ignore =
+    D10,
+    D202,
+    D203,
+    D213,
+    D401,
+    D402,
+    D405,
+    D413,
+    D415,
+    D416,
+match-dir = '[^\._].*'
+{% endblock %}
+
 {% block pylint %}
 [pylint]
 # note: pylint doesn't look in setup.cfg by default, need to call it with
 # `pylint ... --rcfile=setup.cfg`
+# Also, this needs to be the last section in this file, so any additional
+# configuration must be added above here.
 {% if pylint.ignore %}
 ignore =
     {% for ignore in pylint.ignore %}
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -100,9 +100,9 @@ manifest_in: {}
 setup_py:
   entry_points:
     console_scripts:
-      - bones-generate=nengo_bones.scripts.generate_bones:main
-      - bones-check=nengo_bones.scripts.check_bones:main
-      - bones-pr-number=nengo_bones.scripts.pr_number:main
+      - bones-generate = nengo_bones.scripts.generate_bones:main
+      - bones-check = nengo_bones.scripts.check_bones:main
+      - bones-pr-number = nengo_bones.scripts.pr_number:main
   include_package_data: True
   # note: docs/tests_req are intentionally empty to test that the correct
   # requirements are built into the ci scripts
@@ -150,6 +150,7 @@ setup_cfg:
   pylint:
     ignore:
       - nengo_bones/tests/ignoreme
+
 
 docs_conf_py:
   exclude_patterns:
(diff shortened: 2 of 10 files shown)


nengo_bones/templates/setup.cfg.template
```diff
@@ -96,9 +96,24 @@ allclose_tolerances =
 {% endfor %}
 {% endif %}
 
+[pydocstyle]
+ignore =
+    D10,
+    D202,
+    D203,
+    D213,
+    D402,
```
hunse: Probably also want `D401`, which is the imperative mood one. Also, maybe we should put comments after each of these lines with a word or two saying what it is.

tbekolay: The commit message of d2ce6d5ffca0eca0adf488384e75505615924733 gives my thoughts on D401

tbekolay: And I'd rather just link to http://www.pydocstyle.org/en/3.0.0/error_codes.html ... we don't gives description for flake8, which is similarly code-based

hunse:
For D301 (backslashes in docstring), the `nengo.Connection` docstring has a backslash for a line continuation, and it actually doesn't compile with Sphinx if you escape it (i.e. start with `r"""`). But, it does look bad when you call `help` on `Connection` because the unescaped backslash makes a ton of space.

nengo_bones/templates/setup.cfg.template
hunse: It does take a bit of work to cross-reference the two lists, but not many people are going to be editing this, so it's not a big deal to me.

hunse: Also, D415 and D416 are missing from that list? They need to update their docs!

tbekolay:
For the `Connection` docstring, we managed to get around that because Black gives us 88 characters to work with instead of 79, so we didn't need the continuation character after all. For other cases of really long types, I think the general strategy should be to describe it more succinctly (there's almost always a way to reword).

nengo_bones/templates/setup.cfg.template
tbekolay: Yeah they actually have a problem with their current docs right now as they don't render the table: https://github.com/PyCQA/pydocstyle/issues/380

hunse:
I also recall there being one place (with nested functions) where Black and pydocstyle were in conflict. `pydocstyle` wants no newline after function docstrings (D202). In this case, though, the function started with the definition of a helper function, and Black wanted a newline before that helper function definition. I can't seem to find where that was, though.

tbekolay:
I ran these checks on nengo bones and nengo core and aside from D401 they pass, did you run it on any other projects?

hunse:
Nope, I haven't done it on others. I think those are all the tricky situations I found, and it looks like they're all resolved.

As for D401, my vote would be to try making the changes on Nengo and see how hard it is to do in practice, and whether it seems like an improvement.

hunse (approved):

tbekolay:
Just a note that I marked this as a draft because the intention isn't to review/merge it now as I don't want to block https://github.com/nengo/nengo/pull/1557 I just found this useful in reviewing https://github.com/nengo/nengo/pull/1557 and made this so we can discuss it in the future (probably around the time of the Bones 1.0 work).

drasmuss:
My vote would be to disable D401. I don't think it really helps with readability (we're all used to parsing sentences where the verb doesn't come first, or isn't imperative), and is just a pain in the nonzero number of cases where writing in non-imperative style is a natural way to document something. But I'm fine if someone wants to try it out 🤷‍♂.

PR #58, opened 2019-08-22
Add setup.cfg options for pytest_nengo

tbekolay:
**Motivation and context:**
https://github.com/nengo/nengo/pull/1566 changes how Nengo core exposes things that Nengo backends need for running Nengo tests. This supports those changes.

**How has this been tested?**
todo

**Types of changes:**

**Checklist:**

**Still to do:**
- [x] Docs
- [x] Changelog entry
- [x] Tests ?
- [x] Can we simplify the pytest section?


tbekolay:
I added docs, changelog entries, and refactored the pytest option handling a bit to simplify things. Should be ready for review now @drasmuss!

nengo_bones/templates/setup.cfg.template
```diff
@@ -85,49 +83,22 @@ markers =
...
-{% if pytest.nengo_test_unsupported %}
-nengo_test_unsupported =
-{% for var, val in pytest.nengo_test_unsupported.items() %}
+  {% elif type == "dict" %}
+{{ param }} =
+    {% for var, val in vals.items() %}
     {{ var }}
         "{{ val | wordwrap(break_long_words=False, wrapstring="\n        ") }}"
+    {% endfor %}
+  {% endif %}
```
drasmuss: This is probably a personal preference thing, but for me, I prefer if the indentation of the control structures (e.g. for/if) matches the indentation of the text being formatted.  So, e.g., I would format this like
``` python
{% for param ... %}
...
{% elif type ... %}
{{ param }} =
    {% for ... %}
    {{ var }}
        "{{ ...
     {% endfor %}
{% endif %}
{% endfor %}
```
You lose the scoping information in the control blocks, but I find it's easier to see what the resulting text would look like. Otherwise when you're scanning each row you have to mentally parse "is this a control-block indent? or a formatting indent?".

nengo_bones/templates/setup.cfg.template
```diff
@@ -67,9 +67,7 @@ max-line-length = 88
 [tool:pytest]
 {% if pytest.addopts %}
 addopts = {%- for opt in pytest.addopts %} {{ opt }} {%- endfor %}
-
```
drasmuss: This newline is necessary (otherwise `norecursedirs` ends up on the same line as `addopts`).

tbekolay: Right yeah will add it back. I wish Jinja's whitespace handling was better

nengo_bones/templates/setup.cfg.template
tbekolay: I did it that way because I kept missing `endfor`s and `endif`s or having them in the wrong order, was originally going to flatten it all but forgot. In any case, now that it's working I'm fine with whatever indentation for the control blocks :+1:

drasmuss:
Fixups look good to me

PR #60, opened 2019-09-27
Add bones-stale-branches command

tbekolay:
This command lists all stale branches in a repository. If you're in a repository with a .nengobones.yml, it will use the provided information. You can also pass a repo on the command line.

Some known issues:

- In order to avoid rate limiting, you must provide a Github username and personal access token as environment variables. This could be done in a nice way, like using the full OAuth v2 workflow.
- The call to the list of branches is not properly paginated,so only the first chunk of branches is fetched properly.

**Motivation and context:**
In looking at https://github.com/nengo/nengo/issues/1485 I wondered if this is the kind of thing we should do regularly (remove stale branches). Since we already have the `bones-pr-number` script, I thought this might be pretty straightforward. Unfortunately it was not, so before putting more time into it, we should figure out if this is something we actually want.

**How has this been tested?**
Ran the command for Nengo Bones and Nengo Core.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -103,6 +103,7 @@ setup_py:
       - bones-generate=nengo_bones.scripts.generate_bones:main
       - bones-check=nengo_bones.scripts.check_bones:main
       - bones-pr-number=nengo_bones.scripts.pr_number:main
+      - bones-stale-branches=nengo_bones.scripts.stale_branches:main
   include_package_data: True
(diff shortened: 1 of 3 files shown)


hunse:
It's straightforward to view stale branches in Github. Is that sufficient?

tbekolay:
The idea with something like this is that we would use it as part of an automated process. But yeah not necessarily convinced it's a process we care about at this point.

PR #66, opened 2019-10-16
Error when file is missing from config

tbekolay:
**Motivation and context:**
A case that we did not anticipate is as follows:

1. A non-templated file contains the text "Automatically generated by nengo-bones"
2. Since it contains that text, we assume it's in the .nengobones.yml config.
3. It's not there because it's non-templated.

That resulted in a KeyError, which isn't very user friendly and means that the rest of the files don't get checked. This PR detects that case, marks the file as having a problem, gives better error text, and continues on.

**How has this been tested?**
Added a test. Refactored the `test_check_bones.py` file in order to write the test as the one test in there was getting too long.

**Types of changes:**

**Checklist:**


nengo_bones/tests/utils.py
```diff
@@ -4,24 +4,6 @@
 from traceback import print_tb
 
 
-def write_file(tmpdir, filename, contents):
```
drasmuss: What about just changing this function to do the snippet repeated in all those tests (i.e. `tmpdir.join(filename).write(dedent(contents))`)? Doesn't make a big difference, but it'd slightly reduce the boilerplate. Could also add logic to handle the leading newline (seems like it'd be easy to forget the `\`).

tbekolay: Personally I find the calls to `write_file` really hard to read because the function parameters are indented the same level as the contents of the file, so it all just blends together. I also like having the boilerplate there in case you need to write a different file in a nested directory, you can tell from looking at how the simple case goes how the more complicated case would go. While there are a few function calls in the boilerplate, it still all fits on the same number of lines (maybe fewer?) which, to me, means it's the same length. It's true that forgetting the `\` is quite possible though. But, since we only use it in tests, the test would just fail and you'd add it. I think part of it for me is also having around that `tests.utils` module, I would really like to get rid of it because it feels like an anti-pattern to me, test files shouldn't depend on other test files.

Those are my thoughts / motivations to remove it, so yeah, I guess it'd be good to get a third opinion.

drasmuss: Yeah I have the opposite impression, I find the `write_file` code easier to read because it's not spread across multiple indentation levels.

You could also do the nested directory easily with the `write_file` util (e.g. `write_file(tmpdir, "a/nested/file", contents)`).

And yeah, I know your passionate crusade against `utils` :wink:. But it feels like we're making the code more complicated just for personal aesthetic reasons in this case.

This just seems like the canonical case of "a thing that should be a function" (repeating the same code snippet over and over with different arguments).  E.g., what if we decided we wanted to change something about how we write all these test files (maybe for some reason it becomes important to us that all files end with a blank line, or we want to place them all in a subdirectory; who knows)? Instead of searching through all our tests for all the places we do this, we would just make one modification in the `write_file` function. A positive code pattern!

But yeah if you want to get a third opinion on this, we can do that, perhaps after the next scrum.

nengo_bones/tests/test_check_bones.py
```diff
@@ -1,24 +1,20 @@
 # pylint: disable=missing-docstring
 
-from textwrap import dedent
-
 from click.testing import CliRunner
 
 from nengo_bones.scripts import check_bones, generate_bones
-from nengo_bones.tests.utils import assert_exit
+from nengo_bones.tests.utils import write_file, assert_exit
```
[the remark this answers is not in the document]
drasmuss: We should make a decision about whether this is a thing we're enforcing or not (I'd thought we were just accepting black's styling and not enforcing additional style issues after that). Doesn't seem worth it to me, but I'll make the fixup :smile:.

nengo_bones/tests/utils.py
tbekolay: > You could also do the nested directory easily with the `write_file` util (e.g. `write_file(tmpdir, "a/nested/file", contents)`

It's not clear to me from the kwargs that you can do that, and with pytest-plt it took way more time than it should have to get tests passing on Windows because of different path separators.

> This just seems like the canonical case of "a thing that should be a function" (repeating the same code snippet over and over with different arguments).

I think part of why I changed it is because the original implementation was just a reimplementation of some path things (which the `tmpdir` fixture handles well) and the `dedent` function. Then when I saw that the `tmpdir` stuff had a `write` method it seemed like a no-brainer to remove code that we didn't need to write ourselves. I find the `tmpdir.join(file).write(...)` part to be really nice, far preferable to me than `write(tmpdir, file, ...)`. But it is true that the `dedent` part is less satisfying due to needing the backslash, so after thinking about it a bit I'm fine with keeping `write_file` but I think I would like to have all of the arguments passed with their keywords specified; especially since they're being split onto separate lines anyway, it will likely alleviate my initial confusion seeing so many things indented to the same level.

drasmuss: Done

PR #70, opened 2019-11-15
Fix cleanup lock by not invoking subshell

tbekolay:
**Motivation and context:**

There seem to be issues when invoking a subshell in an SSH command in some environments, resulting in Travis CI failing with

```
wait: pid <pid> is not a child of this shell
```

The sequence of `flock` commands used in this commit should be equivalent to the subshell version, but it does not use a subshell.

See [this NengoDL job](https://travis-ci.org/nengo/nengo-dl/jobs/612476553#L1157)

**How has this been tested?**
Not heavily tested yet, we should have NengoDL use this branch and see if the issue is fixed.

**Types of changes:**

**Checklist:**


tbekolay:
This appears to have been a false alarm! It wasn't the subshell causing the issue :crossed_fingers:

PR #78, opened 2020-01-06
Remove coverage.py early starting logic

drasmuss:
**Motivation and context:**

We were using an odd approach to try to early-start the coverage engine to help work around the issues caused by Nengo being imported before coverage data starts being collected.  However, this causes some issues in the new coverage.py 5.0 release (related to this issue https://github.com/nedbat/coveragepy/issues/883).  Fortunately, since we reworked the way Nengo's pytest infrastructure works in Nengo 3.0 this is no longer necessary, so I just removed it rather than trying to fix it.

Note that this may mean that coverage is not collected correctly when testing against Nengo<3.0. However, as long as those jobs are being combined with some coverage jobs running in Nengo>=3.0, I believe the overall coverage information should still be correct.

**How has this been tested?**

Ran into this in https://github.com/nengo/nengo/pull/1591, so I will check that the problem is solved there. I'll also check nengo-dl and nengo-loihi, to make sure that the coverage is still correct there after removing this logic.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,6 @@ env:
   global:
     - SCRIPT="test"
     - TEST_ARGS=""
-    - COV_CORE_SOURCE=nengo_bones  # early start pytest-cov engine
-    - COV_CORE_CONFIG=.coveragerc
-    - COV_CORE_DATAFILE=.coverage.eager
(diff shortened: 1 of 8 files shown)


drasmuss:
Confirmed that this change doesn't affect coverage in in nengo, nengo-dl, and nengo-loihi, so I think this is ready to go.

PR #97, opened 2020-03-31
Install nengo-bones and nengo-sphinx-theme master

drasmuss:
**Motivation and context:**

In the current system we need to do a `nengo-bones` or `nengo-sphinx-theme` release every time we want to make a change in those repos that we want to distribute to downstream repos. This is kind of cumbersome, so instead we'll install from `master` instead.

**How has this been tested?**

Tested as part of the CI in this repo

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Automatically update downstream repos during CI

### DIFF
--- a/.templates/auto-update.sh.template
+++ b/.templates/auto-update.sh.template
@@ -1,0 +1,25 @@
+{% extends "templates/base_script.sh.template" %}
+
+{% block script %}
+    COMMIT_HASH="$(git log -1 --pretty=%h)"
+    COMMIT_MSG="$(git log -3 --pretty='%h - %s')"
+    git config --global user.email "info@appliedbrainresearch.com"
+    git config --global user.name "Nengo Bones"
+
+    {% for repo in repos %}
+    # Update {{ repo }}
+    exe git clone "https://$GH_TOKEN@github.com/{{ repo }}" {{ repo }}
+    cd {{ repo }} || exit
+    exe bones-generate
+    if git diff HEAD --exit-code; then
+        echo "No changes to {{ repo }}"
+    else
+        exe git add --all
+        exe git commit -m "Automatic update to Nengo Bones $COMMIT_HASH" -m "$COMMIT_MSG"
+        exe git push -fq "https://$GH_TOKEN@github.com/{{ repo }}" master:bones-autoupdate
+        echo "Updated {{ repo }}:bones-autoupdate"
+    fi
+    cd .. || exit
+
+    {% endfor %}
+{% endblock %}
--- a/docs/examples/configuration.ipynb
+++ b/docs/examples/configuration.ipynb
@@ -841,6 +841,43 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "### Slack notifications\n",
+    "\n",
+    "The ``slack_notifications`` config option can be used to enable Slack notifications on failed builds. This uses a secret string, which can be generated using the Travis CI command line tool\n",
+    "\n",
+    "```text\n",
+    "travis encrypt <account>:<token>#<slack-channel>\n",
+    "```\n",
+    "\n",
+    "Where `account` and `token` can be obtained from Slack's Travis CI integration settings, and `slack-channel` is where we want messages to be delivered. This will output a string that looks like\n",
+    "\n",
+    "```text\n",
+    "secure: \"<secret-key>\"\n",
+    "```\n",
+    "\n",
+    "Copy that secret key and set it as the value of ``slack_notifications``:"
+   ]
+  },
+  {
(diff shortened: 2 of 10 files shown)


.templates/auto-update.sh.template
```diff
@@ -0,0 +1,21 @@
+{% extends "templates/base_script.sh.template" %}
+
+{% block script %}
+    COMMIT_MSG="$(git log -1 --pretty=%B)"
```
hunse, line 4: This shows the full body for the last commit. Another option is to show the commit hashes and subjects for something like the last 3 commits (`git log -3 --pretty='%h - %s'`). I'm thinking that a) some PRs, like this one, have more than one commit, and b) we don't necessarily need the full body from the Nengo Bones commit in every downstream repo.

.templates/auto-update.sh.template
```diff
@@ -0,0 +1,21 @@
...
+
+    {% for repo in repos %}
+    exe git clone "https://$GH_TOKEN@github.com/{{ repo }}" {{ repo }}
+    cd {{ repo }} || exit
+    exe bones-generate
+    if git diff HEAD --exit-code; then
+        echo "No changes"
+    else
+        exe git add --all
+        exe git commit -m "Automatic Nengo Bones update" -m "$COMMIT_MSG"
```
hunse, line 16: I'm thinking we should change this to `git commit -m "Automatic update to Nengo Bones $COMMIT_HASH" -m "$COMMIT_MSG"`, where `COMMIT_HASH` is the abbreviated commit hash of the latest Nengo Bones commit. This would be a bit more in line with our current way of naming the updates (with "Update to Nengo Bones v0.0.0"), and makes it easier to tell when looking at an abbreviated log where it's up to date with.

docs/examples/configuration.ipynb
```diff
@@ -1819,9 +1855,22 @@
...
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
    "name": "python",
-   "pygments_lexer": "ipython3"
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.6.10"
```
hunse, line 1873: Do we have a spec for what we do with this metadata? I thought that we were trying to keep it out of notebooks, and I typically use the clear-notebooks script to get rid of it. It really doesn't matter either way, just wondering if we have a policy.

.templates/auto-update.sh.template
drasmuss: I suspect anyone who really wants to know what has changed is just going to go look at the Nengo Bones commit history/changelog, so I don't think it matters too much. This was just kind of "well it's easy to do and doesn't hurt, so might as well". But `git log -3 --pretty="%h - %s"` is equally easy, so doesn't matter to me.

hunse:
I've put my suggestions for discussion in a commit 3582f5d.

Also, I was testing the auto-update script locally, and it changed my Git name to "Nengo Bones". This is what we want when we're running on TravisCI, and we don't really expect people to run it locally, and I don't see an easy fix. Just funny, is all!

.templates/auto-update.sh.template
drasmuss: Knowing the commit hash wouldn't be very useful information to me. Like I just can't think of any scenario where I'd be looking at the commit hash in the `nengo/nengo` git history and trying to cross-reference that with the `nengo/nengo-bones` git history. So my slight preference is to avoid it just because it makes the commit history a bit messier, but if other people do have workflows where that'd be useful information to them then I'm fine with it.

docs/examples/configuration.ipynb
drasmuss: Ah yeah I must have just committed this by accident at some point and didn't notice.

.templates/auto-update.sh.template
hunse: I have a `git lg` alias that does a one-line-per-commit summary of the log, so it'd be nice for me to be able to see the abbreviated hash there e.g. if I want to quickly double-check that a repo is up to date with the latest NengoBones. Also, it makes it possible to differentiate between the auto-update commits just from my `git lg` view. Because I often use that view, I really try to avoid commits with the same subject line.

docs/examples/configuration.ipynb
hunse: Ok, undid it.

hunse:
When the configuration notebook was rendered in the docs, it was applying syntax highlighting to the `travis encrypt` block, which was making `#<slack-channel>` look like a comment. I fixed this by specifying "text" formatting for those blocks.

PR #115, opened 2020-09-02
Add support for 404 page in docs

drasmuss:
**Motivation and context:**

Adds support for 404 pages in `nengo-sphinx-theme`.

Supports https://github.com/nengo/nengo-sphinx-theme/pull/65

**How has this been tested?**

Tested in nengo-sphinx-theme by building the docs with these options. There are no configuration options or anything like that, so I don't think we need any tests/docs/changelog in nengo-bones.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo_bones/templates/docs.sh.template
+++ b/nengo_bones/templates/docs.sh.template
@@ -5,7 +5,7 @@
 {% block install %}
 {{ super() }}
     exe pip install \
-        "sphinx>=1.8.0,<3.1.0" \
+        "sphinx>=3.0.0,<3.1.0" \
         "jupyter>=1.0.0" \
         "nbsphinx>=0.2.13" \
         "numpydoc>=0.6.0" \
--- a/nengo_bones/templates/docs/conf.py.template
+++ b/nengo_bones/templates/docs/conf.py.template
@@ -21,6 +21,7 @@ extensions = [
     {% if html_redirects %}
     "nengo_sphinx_theme.ext.redirects",
     {% endif %}
+    "notfound.extension",
     "numpydoc",
     {% for ext in extensions %}
     "{{ ext }}",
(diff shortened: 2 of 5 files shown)


hunse (approved):
LGTM. The docs URL is now based off the `repo_name`, and I tested that it works for NengoBones, NengoOCL, and abr/lmu (which covers repos in nengo, nengo-labs, and abr). I've tested the 404 page with the NengoOCL docs, and it seems to work well: https://labs.nengo.ai/nengo-ocl/test.html

drasmuss:
Added a fixup to simplify things a bit, removing the need for the new `docs_url`/`notfound_urls_prefix` config options. LGTM

nengo_bones/templates/docs/conf.py.template
```diff
@@ -61,7 +61,7 @@ nbsphinx_timeout = -1
 
 # -- notfound.extension
 notfound_template = "404.html"
-notfound_urls_prefix = "{{ notfound_urls_prefix }}"
+notfound_urls_prefix = "/{{ pkg_name | replace("_", "-") }}/"
```
tbekolay, line 64: Safer to use `repo_name` and split it on `/` as this will fail for, e.g., when we do `nengo/nengo-core` (which will still have `nengo` as the `pkg_name`

drasmuss: Wouldn't the documentation still be at `nengo.ai/nengo`?

tbekolay: No, the location of the documentation is based on the repository's name, so if we move the repo to `nengo/nengo-core` (which I think is the plan) then it will move all the docs to `nengo.ai/nengo-core`. Which means that we should definitely put a bunch of redirects in the `nengo/nengo` metapackage before we actually do the move.

nengo_bones/templates/docs.sh.template
```diff
@@ -9,6 +9,7 @@
         "jupyter>=1.0.0" \
         "nbsphinx>=0.2.13" \
         "numpydoc>=0.6.0" \
+        "sphinx-notfound-page>=0.5.0" \
```
hunse, line 12: Could/should we put this in `setup.py` for nengo-sphinx-theme? Perhaps in the "optional" section, then we can install with `[optional]` below.

drasmuss: Ah yeah it should be there. I'd just put it in the main requirements, as I think we'd expect/want people to be using the 404 page extension in almost all cases (I can't think of any reason why you wouldn't), and then people don't need to remember the `[optional]` tag.

hunse:
Merged

PR #116, opened 2020-09-03
Apply prettier before text wrapping

hunse:
This makes sure long lines (e.g. links) from prettier will still
be wrapped, and thus the static checks will pass.


tbekolay:
Looks like this got included in a recently merged PR: cb41790a8ef1455528a680f8104de901eec4bf23

PR #118, opened 2020-09-03
Fix small typos

tbekolay:
Some tiny things I saw looking at the `bones-format-notebook` diff.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -82,7 +82,7 @@ ci_scripts:
       - npm install -g npm stable
       - npm install -g node --force
       - node --version
-      # install prettier for use in formate_notebook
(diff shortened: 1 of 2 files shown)


hunse (approved):

hunse:
Looks good, but some of the tests are failing...

tbekolay:
> Looks good, but some of the tests are failing...

Ah... so this happens because doing `passed = passed and ...` short-circuits, meaning that if a previous check had failed, it doesn't run the other check. A noble optimization but we want to run it regardless. I could switch the order of the checks so that `apply_static_checker` goes first, but Black doesn't format that too well, so I'll revert those back to `&=`.

PR #121, opened 2020-09-24
Add isort configuration

tbekolay:
This makes it possible to run `isort {{ pkg_name }}` inside projects that use NengoBones, and will result in correct behavior if you set up your editor to run `isort` when saving a file.

This is the first step in automating the sorting of our imports. I opted to use the "black" profile, since we format with black, but aside from that opted for the default behavior of isort.

**Motivation and context:**
I was annoyed sorting imports in `abr/tools`, so I decided to see how quick it would be to set up `isort`. It was pretty quick.

**How has this been tested?**
Ran `isort nengo_bones` inside this repo. See the diff of this PR for the results.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Add `isort --check` to the `static.sh` template
- [x] Add changelog entry
- [x] Add docs, test? Probably just docs?
- [ ] Confirm everyone's cool with how it sorts imports


nengo_bones/tests/test_format_notebook.py
```diff
@@ -4,10 +4,10 @@
 import re
 import sys
 
-from click.testing import CliRunner
-from nbconvert.preprocessors import ExecutePreprocessor
 import nbformat
 import pytest
+from click.testing import CliRunner
+from nbconvert.preprocessors import ExecutePreprocessor
```
hunse, line 10: So all `from` statements have to come after `import` statements in a block? This is not how we do things right now. Is there a profile/setting that would be more similar to what we do now (both to reduce changes, and because I like what we do now)?

tbekolay: There are configuration options, yes, I would be surprised if that wasn't configurable: https://pycqa.github.io/isort/docs/configuration/options/

I personally don't have strong opinions on what to do, as I would rather leave it up to `isort`. I would imagine that, over time, people would standardize on whatever the default option is, so I didn't spend any time looking through the configuration options to make it look like what we have now. The size of the diff is not a major factor as I imagine we would have all the changes in one commit that runs isort on anything anyhow. So yeah, I feel like we would need a pretty strong reason to opt for a particular configuration over the default.

tbekolay:
Since I have isort set up with my editor, it's now getting annoying having to remember to turn off my pre-save hook when working with Nengo repos, so I finished up the remaining tasks for this PR and would like to get it merged ASAP!

drasmuss (approved):
Looks good, just made a minor change mentioning the fact that `isort` will be checked during static checks to the changelog (squashed it in so we don't have to wait any longer for CI).

PR #124, opened 2020-10-29
Add remote-script template

drasmuss:
**Motivation and context:**

Something that has come up a lot is that we want to run one of the CI scripts on some remote machine. This script builds on top of the more general `remote.sh` template to make that specific use case easier.

NengoDL, KerasSpiking, and NengoFPGA all currently have their own versions of this script; we should switch them over to using this template (and make sure that this script covers all those use cases).

**How has this been tested?**

Added some tests on the content of the rendered script, as well as a new functional job which runs the test script on a remote.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo_bones/templates/remote-script.sh.template
+++ b/nengo_bones/templates/remote-script.sh.template
@@ -1,0 +1,58 @@
+{% extends "templates/remote.sh.template" %}
+
+{% block remote_install %}
+{{ super() }}
+        # copy environment variables into remote environment
+        export TRAVIS_JOB_NUMBER="$TRAVIS_JOB_NUMBER"
+        export TRAVIS_BRANCH="$TRAVIS_BRANCH"
+        export TRAVIS_TAG="$TRAVIS_TAG"
+        export TEST_ARGS="$TEST_ARGS"
+        export PIP_USE_FEATURE="$PIP_USE_FEATURE"
+        {% if remote_vars %}
+        {% for var, val in remote_vars.items() %}
+        export {{ var }}="{{ val }}"
+        {% endfor %}
+        {% endif %}
+
+        # generate CI scripts on remote device
+        # ensure nengo-bones version is the same as the one installed on host
+        pip install $(pip freeze | grep nengo-bones | awk '{print $(NF)}')
+        bones-generate --output-dir .ci ci-scripts || REMOTE_STATUS=1
+
+        {% for line in remote_setup %}
+        {{ line }}
+        {% endfor %}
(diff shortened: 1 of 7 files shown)


nengo_bones/templates/remote-script.sh.template
```diff
@@ -0,0 +1,74 @@
...
+
+        echo "Waiting for lock on device $DEVICE_ID"
+        (
+            flock -x -w 540 200 || exit 1
+            bash .ci/{{ remote_script }}.sh script || exit 1
+        ) 200>/var/lock/.travis-ci.exclusivelock."$DEVICE_ID" || REMOTE_STATUS=1
+{% endblock %}
+
+{% block after_script %}
+    {% if remote_script == "docs" %}
```
hunse, line 64: Just a personal preference thing, but I would un-indent these by two spaces just to make it easier to distinguish the templating from the actual code in this block.

tbekolay: I'm pretty indifferent (the Jinja templates don't have a well-specified style) but I'm going to leave it as is to be consistent with the other templates in this repo.

tbekolay (approved):
I made a small fixup to replace the `awk` command with `cut`. The only other thought that I had was that a bunch of things will break if a CI script has a space in it, which we could add a check for somewhere if we're worried about it. I'm not worried for the repos we control now, but new repos/teams/external users could get tripped up with it :shrug:

Anyway if the fixup looks good to you @drasmuss I'll squash and merge.

Update: to speed things up (travis-ci.org is super slow) I did the squash already; the line I changed was [this one](https://github.com/nengo/nengo-bones/pull/124/commits/e7c63df022baa9ea5ff5453c6f8bdd75489c3c45#diff-aeeba8e1bc1b36bc682c34f7e0d550c433759668a66dfcd9d290177812600a8fR19)

drasmuss:
I used awk because it allows us to get the last element in a series of unknown length, because I'm not sure that it'll always be length 2. E.g., when doing pip freeze I've seen nengo-bones @ git+https:..., but also just -e git+https://. It's probably relatively reliable here since we're always using the same environment and installing in the same way, but it doesn't hurt to be careful I think unless you have a strong reason for avoiding awk.

tbekolay:
> I'm not sure that it'll always be length 2

Ah, didn't realize that the `pip freeze` lines could be different. Generally I find `awk` commands inscrutable but `cut` can't get the last field reliably so I'll revert :+1:

PR #130, opened 2020-11-27
Use TravisCI.com

hunse:
**Motivation and context:**

Migrate to using TravisCI.com for builds.

**How has this been tested?**

Checked that the build is running on travisci.com.

Retrying the miniconda install in the script has been successful in at least one build: https://travis-ci.com/github/nengo/nengo-bones/jobs/449899508

**Types of changes:**

**Checklist:**

- [ ] I have run the test suite locally and all tests passed.


nengo_bones/templates/remote.sh.template
```diff
@@ -93,12 +93,36 @@ See https://docs.travis-ci.com/user/encrypting-files/ for more details.
...
+        DOWNLOAD_STATUS=0
+        wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh --quiet -O miniconda.sh || DOWNLOAD_STATUS=1
+        if [ "\$DOWNLOAD_STATUS" -gt 0 ]; then
+            echo "({{ host }}) Failed to download miniconda. Exiting.";
+            exit \$DOWNLOAD_STATUS;
+        fi
+        for i in \$(seq 1 10); do
+            echo "$ ({{ host }}) Installing miniconda (try \$i)"
+            INSTALL_STATUS=0
+            bash miniconda.sh -b -p ./miniconda || INSTALL_STATUS=1
```
drasmuss, line 106: If the `./miniconda` directory already exists (e.g., because we're re-running a previous build) then this will fail, which we don't want to treat as a critical error. We could either delete the `./miniconda` directory ahead of time if it exists, or we could make this whole download/install block conditional on the `./miniconda` directory not existing. I'd probably lean towards just deleting it if it exists, so that we're starting from a clean slate each time.

nengo_bones/templates/remote.sh.template
```diff
@@ -93,12 +93,36 @@ See https://docs.travis-ci.com/user/encrypting-files/ for more details.
...
-        wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh --quiet -O miniconda.sh
-        bash miniconda.sh -b -p ./miniconda
+        echo "$ ({{ host }}) Downloading miniconda"
+        DOWNLOAD_STATUS=0
+        wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh --quiet -O miniconda.sh || DOWNLOAD_STATUS=1
+        if [ "\$DOWNLOAD_STATUS" -gt 0 ]; then
+            echo "({{ host }}) Failed to download miniconda. Exiting.";
+            exit \$DOWNLOAD_STATUS;
+        fi
+        for i in \$(seq 1 10); do
```
drasmuss: I don't know if the potential benefit of an installation failing the first time but succeeding on a later attempt is worth all the added complexity here. It's already rare that the installation fails, and it'd be even more rare that it would succeed on a later attempt when that does occur. I'd be inclined to just try the installation once, and exit on failure. On the rare occasions that that happens, people can always just restart the build to try again.

hunse: It's become much more common that the installation fails (see many of the last builds on TravisCI.org, as well as at least one of the new ones on TravisCI.com). I believe it may be because of a flaky internet connection where our machine is hosted. Restarting automatically saves considerable time and hassle.

Here's one example of the restarting working successfully: https://travis-ci.com/github/nengo/nengo-bones/jobs/449899508

nengo_bones/templates/remote.sh.template
hunse: I do delete it if it fails, see below.

drasmuss: Yes but if the `.miniconda` installation succeeds, but then a unit test fails or something like that so you restart the build, then the `.miniconda` directory will still be there.

nengo_bones/templates/remote.sh.template
drasmuss: It's gotta be an `abrgl` issue, because I've literally never seen it fail in NengoDL (which is running on Azure instead). But yeah if it's happening enough to be noticeable on `abrgl` then I guess this is worth it.

nengo_bones/templates/remote.sh.template
hunse: Ahh, I see what you mean. That sounds like a good idea.

drasmuss: I guess after it failed the first time (because the `.miniconda` directory exists), then it would delete it and then try again, so that kind of solves the problem 🤷.

hunse: Ok, I've added a fixup.

drasmuss:
Added some fixups to simplify the retry logic a little bit, if those look good to you, will merge!

hunse:
They look fine. The removal of the download failure message means it's a little less clear if that fails, but a) I've never seen that fail, and b) it should still be possible to figure out exactly where it failed. So I'm good with it.

drasmuss:
Yeah I was assuming that `wget` would already be displaying some kind of failure information if things went wrong

PR #132, opened 2020-11-30
Fix codecov missing files with same names

hunse:
**Motivation and context:**

This fixes a problem with CodeCov not reporting on files where a file in a deeper folder shares the same name. This problem was first observed in Nengo. It has been reported before [here](https://community.codecov.io/t/missing-files-in-codecov-but-not-in-report/825/8); that post also motivates this fix.

There is also the potential of fixing this by using [CodeCov's path fixing](https://docs.codecov.io/docs/fixing-paths). However, I wasn't able to get this to work using the type of fix recommended by tom in [the above thread](https://community.codecov.io/t/missing-files-in-codecov-but-not-in-report/825/8) (specifically, by adding ``fixes:\n    - "::nengo/"`` to `.codecov.yml`).

Based on #131.

**How has this been tested?**

Tested for Nengo in nengo/nengo#1649.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -56,6 +56,8 @@ Release History
 
 - Intermittent failures installing miniconda in the remote script have been fixed by
   retrying the miniconda install if it fails. (`#130`_)
+- Fixed an issue in which Codecov would not report on files where a file in a
+  deeper folder shares the same name. (`#132`_)
 
 .. _#32: https://github.com/nengo/nengo-bones/pull/32
 .. _#101: https://github.com/nengo/nengo-bones/pull/101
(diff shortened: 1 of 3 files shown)


tbekolay (approved):
From looking at how Codecov does things, this seem like the right fix :+1:

IMO the changelog isn't the place to document and contextualize a fix, a changelog entry should be 2 lines maximum just stating the end effect (similar the title of a PR, not the description). We want it quick and scannable, not overwhelming; if people want more context they should look in the PR (or the commit message if offline). I've reworded the changelog message, moved it to the Fixed category, and added more context to the commit message.

Merging!

PR #138, opened 2021-01-11
Configure codespell in setup.cfg

drasmuss:
**Motivation and context:**

This allows configuration options to apply to all invocations of codespell, not just in static.sh. In particular, this is useful if we want to apply ``ignore_words`` to the codespell invocations inside ``bones-format-notebook``.

Note that all downstream repos will need to update their configs if they are currently specifying ``codespell_ignore_words` in the static script config section.  We could technically do this automatically with some kind of deprecation warning, but since we control all the downstream repos that seemed like overkill.

**How has this been tested?**

Covered by existing tests, and added a check that this now works in ``bones-format-notebook``.

**Types of changes:**

**Checklist:**


hunse (approved):
Looks good. I just changed setup.py to also require codespell >= 2.0.0.

PR #139, opened 2021-01-14
Fix version string

tbekolay:
**Motivation and context:**
I did a `pip list` and noticed that the NengoBones version string was wrong. We were outputting versions like

   21.1.dev{dev}

Turns out, we got a bit overzealous with our f-strings and made something pretty inscrutable so we missed an `f` in the nested string.

With this PR the version is now correctly

   21.1.dev0

and does this without nested f-strings, which are hard to read.

**How has this been tested?**
Redid `pip list` and it looks correct now. Also checked by importing and printing `nengo_bones.version.version`.

**Types of changes:**

**Checklist:**

- [na] I have run the test suite locally and all tests passed.

### DIFF
--- a/nengo_bones/version.py
+++ b/nengo_bones/version.py
@@ -13,6 +13,6 @@ today = date.today()
 version_info = (today.year - 2000, today.month)
 dev = 0
 
-version = (
-    f'{".".join(str(v) for v in version_info)}{".dev{dev}" if dev is not None else ""}'
+version = ".".join(str(v) for v in version_info) + (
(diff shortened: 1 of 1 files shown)


hunse:
I added brackets around the `if dev is not None` part (`+` has precedence over `if`). On the one hand, the fact that we've messed it up so many times suggests that we really should have a test. But I do think we've got it this time, so it's probably OK.

tbekolay:
> `+` has precedence over `if`

Good catch, I never remember this! Thanks for the fixup :+1:

PR #140, opened 2021-01-27
Add support for typed projects

tbekolay:
**Motivation and context:**

A few internal projects are now using types. I've been using this branch for those projects, but other people are going to have to be running `mypy` checks soon too, so I fixed up this branch so that the type support is optional. You can opt-in to support by enabling the `py_typed` section, which does two things:

- Adds a {{ pkg_name }}/py.typed file, which lets other packages know that the project is typed.
- Adds mypy checking to static script and pre-commit hook.

The py.typed file is the first templated file we've added that should reside in the package directory. It required a little bit of special case code; specifically, we treat the `templates/pkg` directory as special when generating the section name (we don't include the `pkg_` part) and when saving the file (we replace the `pkg/` part of the path with `{{ pkg_name }}`).

**How has this been tested?**
This is tested in the `test_generate_bones.test_generate_all` test, which required a small fix.

**Types of changes:**

**Checklist:**

**Still to do:**
- [x] Add to config notebook
- [x] Changelog entry
- [x] [Do we need to add `py.typed` to `setup.py` too?](https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages)
- [x] Discuss: does this special casing for the package name make sense?

### DIFF
--- a/nengo_bones/templates/.pre-commit-config.yaml.template
+++ b/nengo_bones/templates/.pre-commit-config.yaml.template
@@ -17,5 +17,12 @@ repos:
 -   repo: https://github.com/pycqa/isort
     rev: 5.6.4
     hooks:
-      - id: isort
-        files: \.py$
+    - id: isort
+      files: \.py$
+{% if py_typed is defined %}
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.800
+    hooks:
+    - id: mypy
+      files: \.py$
+{% endif %}
--- a/nengo_bones/templates/setup.cfg.template
+++ b/nengo_bones/templates/setup.cfg.template
@@ -181,3 +181,10 @@ skip = ./build,*/_build,*-checkpoint.ipynb,./.eggs,./.git,*/_vendor,{{ codespell
 ignore-words-list = {{ codespell.ignore_words | join(",") }}
 {% endif %}
 {% endblock %}
+{%- if py_typed is defined %}
+
+{% block mypy %}
+[mypy]
+python_version = 3.6
+{% endblock %}
+{% endif %}
--- a/nengo_bones/templates/setup.cfg.template
+++ b/nengo_bones/templates/setup.cfg.template
@@ -185,6 +185,6 @@ ignore-words-list = {{ codespell.ignore_words | join(",") }}
 
 {% block mypy %}
 [mypy]
-python_version = 3.6
+python_version = {{ min_python }}
 {% endblock %}
 {% endif %}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,5 +14,5 @@ repos:
 -   repo: https://github.com/pycqa/isort
     rev: 5.6.4
     hooks:
-      - id: isort
-        files: \.py$
+    - id: isort
+      files: \.py$
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,8 @@ Release History
 - Added ``skip`` option to codespell config, which can be used to specify files that
   should be ignored. (`#138`_)
 - Added support for projects with type hints through the ``py_typed`` section. (`#140`_)
+- Added ``min_python`` option to the main section. The default ``python_requires``
+  in ``setup.py`` is now based on this, if not overridden. (`#140`_)
 
 **Changed**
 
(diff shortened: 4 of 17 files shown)


nengo_bones/templates/.pre-commit-config.yaml.template
```diff
@@ -17,5 +17,12 @@ repos:
 -   repo: https://github.com/pycqa/isort
     rev: 5.6.4
     hooks:
-      - id: isort
-        files: \.py$
+    - id: isort
+      files: \.py$
+{% if py_typed %}
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.720
```
hunse, line 24: Why is this 0.720, but in setup.py we require 0.800? (Also, wacky version numbers.)

nengo_bones/templates/setup.cfg.template
```diff
@@ -181,3 +181,10 @@ skip = ./build,*/_build,*-checkpoint.ipynb,./.eggs,./.git,*/_vendor,{{ codespell
 ignore-words-list = {{ codespell.ignore_words | join(",") }}
 {% endif %}
 {% endblock %}
+
+{%- if py_typed %}
+{% block mypy %}
+[mypy]
+python_version = 3.6
```
hunse, line 188: At some point, we might want to add a `min_python_version` variable which we can use to set both this and the `python_requires` in `setup.py`. (I assume that a project should always have this as the minimum version it supports?) This doesn't have to happen in this PR, just something that would be nice for the future.

tbekolay: I wouldn't mind doing it now actually because in the projects that use typing we actually have a minimum version of 3.8. So I'll add that in now.

nengo_bones/templates/.pre-commit-config.yaml.template
tbekolay: Ah just time passing from when I made the first version of this PR and the cleaned up one, haha. I'll update this.

tbekolay:
Pushed a squash commit fixing everything except for the minimum Python version ... but I gotta bounce right now, so I'll add that in this afternoon.

nengo_bones/templates/setup.cfg.template
tbekolay: This is added now!

hunse:
@tbekolay So the decision on including `py.typed` in `setup.py` is that it's not necessary?

tbekolay:
> So the decision on including `py.typed` in `setup.py` is that it's not necessary?

Yes, what I've done instead is added it to `MANIFEST.in` instead. This works as long as `include_package_data=True` in `setup`, and it works for both source and binary distributions. Adding it to `package_data` apparently only works for one of those two, though I forget which. In any case, I did test it by adding `py_typed: {}` to NengoBones, doing `bones-generate` and `python setup.py sdist bdist_wheel`, and then confirmed that the `py.typed` file was in the correct location in both.

I also made a story (for later, not now) to make `include_package_data` default to True as that is a bit of a gotcha for this implementation.

hunse:
I added a couple lines to the format-notebook test, so that the widgets removal is covered.

For some reason, the spacing before the `[mypy]` section wasn't working properly in `setup.cfg` (it was appearing directly after the previous section, with no newlines; adding in newlines to the template between the previous block and the `if` did nothing). Putting the `if` inside the template block seems to have fixed it. I can't think of a downside to this, but I'll let you give the final OK @tbekolay before I merge.

tbekolay:
> Putting the `if` inside the template block seems to have fixed it.

Are you sure that's what fixed it and no moving the blank line to inside the block? I'd prefer the block inside the if if possible.

hunse:
So putting the `if` inside is what fixed it for the case that the `[mypy]` block is present, but then it was adding an extra line when it's absent. I think moving the line inside can work either way, so if you prefer the block inside the `if`, then I think we can switch it back. Can I ask why you prefer it that way?

tbekolay:
> I think moving the line inside can work either way, so if you prefer the block inside the `if`, then I think we can switch it back. Can I ask why you prefer it that way?

Yes that'd be preferred! Blocks are jinja's way of making something easy to extend, so if some other project extended the setup.cfg template to add in some extra stuff to the `mypy` section, they are unlikely to remember to put in the if statement. While that's unlikely to cause an issue (they're overwriting it because they use it) it still makes a mistake possible that wouldn't be possible if the if is outside the block.

hunse (approved):

PR #144, opened 2021-03-03
Add all sections to CI script config

tbekolay:
**Motivation and context:**
In running the static checks in a typed project, I noticed that the `mypy` command was not being run, even though other `mypy` setup was being done correctly in other templates. As it turns out, CI scripts are generated slightly differently compared to other templates. Previously, they did not have access to the rest of the .nengobones.yml file, only their own section, `pkg_name`, and `repo_name`.

This PR makes them more similar to other templates by giving them access to the whole configuration file, with the same promotion of that specific script's information. It fixes the original issue; `mypy` is now run through `static.sh`.

**How has this been tested?**
Tested in downstream typed repo.

**Types of changes:**

**Checklist:**


drasmuss:
Added a minor fixup to make things marginally more snazzy. Is there a quick test that could be added that would catch whatever issue you were having with the `mypy` check not happening? I'm just assuming that my fixup is equivalent, but don't have a way to test that.

tbekolay:
Fixup looks good to me (so fancy!) and I pushed a fixup to test the change. It tests exactly the issue I was having and should fail on master (you can confirm that if you want).

drasmuss (approved):
Looks good

PR #145, opened 2021-03-04
Add main_branch option to main section

tbekolay:
**Motivation and context:**

This option allows repos to use `main` (or anything else) as their default git branch name, instead of the now discouraged `master`. We still set the default value to `master` in order to minimize the number of changes needed to downstream repos, but eventually we should switch this to default to `main`.

This is needed for more recent repos created through TravisCI and the `git init` command.

**How has this been tested?**
Modified `test_load_config` to ensure default value is loaded; ensured that no generated files have changed in this repo, but that they are correct in a downstream repo.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -1,6 +1,7 @@
 project_name: NengoBones
 pkg_name: nengo_bones
 repo_name: nengo/nengo-bones
+main_branch: master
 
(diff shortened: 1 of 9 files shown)


drasmuss (approved):
Looks good

PR #151, opened 2021-05-05
Add version.py template (v2)

drasmuss:
**Motivation and context:**

The content of `version.py` is repeated in all of our repos, so we should template it.

This is based on #39 , but so much has changed since that PR that it was easier to just start fresh.

The main difference functionally from #39 is that we don't internally generate the version string to be used statically in other files (e.g. setup.py and docs/conf.py). The version in those files is still determined dynamically by importing `version.py`. There were two reasons for this.

1. In this PR templating `version.py` isn't mandatory, so we can't count on the static version info being available. However, we could work around this by having conditional logic that uses a static version if `version.py` is templated otherwise falls back to the dynamic import.

2. This PR also supports calendar versioning, which complicates the generation of a static version. The `bones-check` during CI would be updating the static version each time, which means that we'd have to manually update the static version in every repo each month or `bones-check` would fail. Again we could work around this by having conditional logic that uses a static version if you're using semantic versioning but falls back to the dynamic version if you're using calendar versioning.

So as mentioned we could work around both of these limitations with some conditional logic. But I just wasn't sure that adding that complexity is worth it, given that we're doing everything dynamically currently and it's working well enough.

**How has this been tested?**

Added a test (and it's used to template version.py in this repo).

**Types of changes:**

**Checklist:**


nengo_bones/templates/pkg/version.py.template
```diff
@@ -0,0 +1,33 @@
...
+name = "{{ pkg_name | replace("_", "-") }}"
+dev = {% if release %}None{% else %}0{% endif %}
+
+
+# use old string formatting, so that this can still run in Python <= 3.5
+# (since this file is parsed in setup.py, before python_requires is applied)
+version = "{v}{dev}".format(
+    v=".".join(str(v) for v in version_info),
+    dev=(".dev%d" % dev) if dev is not None else "",
+)
```
hunse, line 33: Could we add the copyright string as used e.g. [here](https://github.com/nengo/nengo/blob/32acd92bc97669beda417e5006b72eb90fde3341/nengo/__init__.py#L53)? That way, we can update that dynamically with the year and don't have to manually do it each year. Could also add the `__license__` string that appears there, too, I guess.

drasmuss: Added the copyright string. I didn't do the `__license__` string since it seemed like kind of a stretch to put that in `version.py`, and it will change very rarely so it doesn't seem as important to template it.

nengo_bones/scripts/format_notebook.py
```diff
@@ -33,11 +33,10 @@ def format_notebook(nb, fname, verbose=False, prettier=None):  # noqa: C901
 
     # --- Remove bad metadata
     # Should pass `nengo/tests/test_examples.py:test_minimal_metadata`
-    if "kernelspec" in nb.metadata:
-        del nb.metadata["kernelspec"]
-
-    if "widgets" in nb.metadata:
-        del nb.metadata["widgets"]
+    badmeta = ("kernelspec", "widgets")
```
tbekolay: Super nitpick but I'm going to change this to a list. I think I was convinced of this by a talk from Alex Gaynor where he said that if you could see the length of collection changing, it should be a list, and if not then it should be a tuple. There's also [this post](https://jtauber.com/blog/2006/04/15/python_tuples_are_not_just_constant_lists/).

nengo_bones/version.py
```diff
@@ -22,3 +22,5 @@
     v=".".join(str(v) for v in version_info),
     dev=(".dev%d" % dev) if dev is not None else "",
 )
+
+copyright = "2018-2021 Applied Brain Research"
```
tbekolay, line 26: Did a quick google of how these should be formatted, and how other Python projects format it and it seems like the consensus is `"Copyright (c) 2018-2021 Applied Brain Research"` so I'm going to use that.

tbekolay (approved):
Made two tiny changes, other than that this LGTM! Will merge on CI passing.

PR #152, opened 2021-06-23
Added SLURM and Conda support

xchoo:
Added SLURM, and alias support for CI scripts. 

- Modified exe function in base_script.sh.template to support aliases
- Modified remote-script.sh.template to perform docs tasks for any script ending in "docs"
- Added SLURM support with slurm-script.sh.template
- SLURM support wraps given commands with SLURM commands using bash aliases
- Note: slurm-script supports aliasing of generic commands, while changes to `exe` function support aliasing of commands called with `exe`
- Added usage instructions for slurm-script.sh.template and remote.sh.template
- Updated secret.tar.enc to use `abrghost` instead of `abrgl`
- Added remote slurm test script as an example

**How has this been tested?**
Tested by running test suite. Also checked that slurm runs are being executed properly.

Start with `slurm-script.sh.template`

**Types of changes:**

**Checklist:**

**Still to do:**


CHANGES.rst
```diff
@@ -45,6 +45,8 @@ Release History
 - Added support for changing the main branch name with the ``main_branch``
   config option. (`#145`_)
 - Added template for ``version.py``. (`#151`_)
+- Added support for ``conda install`` in CI scripts (`#152`_)
```
tbekolay, line 48: Period at the end of these entries please.

nengo_bones/templates/base_script.sh.template
```diff
@@ -25,12 +25,16 @@ fi
 
 if [[ "$COMMAND" == "install" ]]; then
     {% block install %}
+    {% if conda_install %}
+    exe conda install {%- for pkg in conda_install %} "{{ pkg }}" {%- endfor %}
```
tbekolay, line 29: I don't think `conda` isn't available to most uses of the `base_script` template. NengoBones is opinionated in some places, and using `pip` is one of them. If you want to use conda you'll have to change more things, and because of that I would force the downstream repo to extend one of the templates and override the `install` block to do `conda` stuff.

xchoo: Oh.. Okay. I'll remove the conda stuff then.

xchoo: What about putting it just in the `remote.sh.template` file? Right before here: https://github.com/nengo/nengo-bones/blob/master/nengo_bones/templates/remote.sh.template#L115 ?

My reason for putting this in is specifically for pandoc. With the other CI scripts, we have the `apt_install` option. But, for remote scripts, doing that is not possible because permissions aren't available. I.e., the only way to have pandoc available is either: get the system administrator to install it, or use conda.

drasmuss: You could just put
```
pre_commands:
    - conda install pandoc
```
in the configuration of your docs script and that would achieve the same goal I believe.

nengo_bones/templates/remote.sh.template
```diff
@@ -112,6 +134,20 @@ See https://docs.travis-ci.com/user/encrypting-files/ for more details.
             echo "({{ host }}) Failed to create conda environment. Exiting.";
             exit \$REMOTE_STATUS;
         fi
+
+        {% if conda_install %}
+        # Install necessary conda packages
+        echo "$ ({{ host }}) Conda installing {%- for conda_pkg in conda_install %} "{{ conda_pkg }}" {%- endfor %}"
+        {% for conda_pkg in conda_install %}
```
drasmuss, line 141: I think this functionality is already supported by the `remote_setup` option in the remote script. E.g., see https://github.com/nengo/nengo-dl/blob/master/.nengobones.yml#L202

nengo_bones/templates/slurm-script.sh.template
```diff
@@ -0,0 +1,76 @@
...
+
+{% block script %}
+    # Enable `expand_aliases` option for bash interpreter
+    shopt -s expand_aliases
+
+    # Create aliases for commands to include SLURM command prefix
+    {% for command in wrapped_commands %}
+    alias {{ command }}="{{ slurm_command }} {{ command }}"
+    echo "Created alias for {{ command }}: "
+    type -a {{ command }}
```
drasmuss, line 47: I don't think you need a new script template for this, I think you could do this with `pre_commands`, same as with the conda installation. E.g., in your docs script you'd put
````
pre_commands:
    - alias sphinx-build="srun sphinx-build"
```

nengo_bones/templates/remote.sh.template
xchoo: Ohhh. This should be documented. 😆

xchoo: Looking into this further, it looks like `remote_setup` is only available if you use the `remote-script` template, which not all repos do, because the actual script itself is highly variable. Putting the `conda install` here would make it more general (or, I could also move the `remote_setup` block into the `remote` template.

nengo_bones/templates/slurm-script.sh.template
xchoo: I will test this

nengo_bones/templates/remote.sh.template
drasmuss: If you're using the `remote.sh` script directly then you have full control over the `remote_script` block, which you could use to do the conda installation. `remote_setup` is only needed if you're using `remote-script.sh`, as then it lets you insert functionality within the `remote_script` block (since `remote-script.sh` is "taking control" of that block).

nengo_bones/templates/slurm-script.sh.template
xchoo: I can't get this to work with `pre_commands`. I think it's the way the aliases work...

xchoo:
This should be final now. I removed the `conda` stuff because it can be done with the `remote_setup` section. However, the SLURM stuff seems to only be doable with `slurm-script`. I tested using `pre_commands` with the NengoFPGA repo, and it wasn't having it. Bash wasn't recognizing the alias, and a lot of other changes (i.e, changes to all of the other scripts) would need to be made to get it to do so.

xchoo:
Okay. I think this is pretty much stable now. I left the `slurm-script` the way it is since it seemed to be the only way generic commands (in the wrapped script) worked with the aliases. The changes to `exe` does give it support for SLURM, so if we can guarantee that every command in the wrapped script is run with `exe`, we can remove `slurm-script`.

PR #153, opened 2021-08-13
Add build_requires option to pyproject.toml

drasmuss:
**Motivation and context:**

This lets projects specify additional build dependencies if needed.

Since it's now possible that a user would want to template `pyproject.toml` but not `pre-commit-config.yaml`, I reworked the validation that used to require both be present if either was present.

Used in https://github.com/nengo/keras-lmu/pull/42, where it is needed to add the `oldest-supported-numpy` dependency.

**How has this been tested?**

Added a test, see `test_generate_bones.py::test_pyproject_toml`.

**Types of changes:**

**Checklist:**

**Still to do:**

- [ ] Update documentation


drasmuss:
This didn't end up fixing the problem I hoped it would fix in https://github.com/nengo/keras-lmu/pull/42, so there's no particular motivation to merge this in now. But it works, so I'll leave it up in case we want to come back to this in the future.

PR #154, opened 2021-09-16
Update version.py template

drasmuss:
**Motivation and context:**

Disable pylint f-string warning (since we explicitly want to use old-style formatting in version.py). 

Also rearrange version string so that mypy doesn't get confused about the type when dev=None (it wasn't correctly parsing the in-line `if` condition to realize that `dev` could never be None in the `.dev%d` case).

**How has this been tested?**

The f-string warning is tested in this repo (by virtue of the static checks passing). I manually verified the typing changes work properly in a repo that uses typing.

**Types of changes:**

**Checklist:**

**Still to do:**


drasmuss:
I very slightly prefer non-zero padded just because it's simpler. From a brief search, calver projects seem pretty evenly split on whether to zero pad or not. But I'm fine with changing it :+1:. 

And I'd just include the day for all our calver projects. It doesn't hurt, and can come in handy if we ever need to release more frequently.

tbekolay:
> From a brief search, calver projects seem pretty evenly split on whether to zero pad or not. But I'm fine with changing it +1.

Probably [this search](https://github.com/search?q=%22__version__+%3D+2021.*%22&type=Code) is more accurate, some semver projects do have 21 major versions. My main bias comes from [Ubuntu](https://wiki.ubuntu.com/Releases) and [youtube-dl](https://github.com/ytdl-org/youtube-dl/releases) but yeah it is a judgment call. But I am going to change it :sunglasses: 

> And I'd just include the day for all our calver projects. It doesn't hurt, and can come in handy if we ever need to release more frequently.

Yeah agreed :+1:

tbekolay (approved):
Fixup is 65039c4b5f0b6a4ef7d827a95b3de10aa9f50e33; didn't want to change how semver works and I thought we might want to keep the `version_info` as ints for consistency. Will squash and merge when CI passes :+1:

PR #155, opened 2021-09-22
Remove leading zeros from calver version string

drasmuss:
**Motivation and context:**

We noticed that the leading zeros get removed by pip (see https://www.python.org/dev/peps/pep-0440/#integer-normalization), so removing them from the version string to make things consistent.

**Types of changes:**

**Checklist:**


tbekolay (approved):
LGTM!

PR #157, opened 2021-09-29
Update license to use ABR license

tbekolay:
**Types of changes:**

**Checklist:**


LICENSE.rst
```diff
@@ -6,23 +6,20 @@ NengoBones license
...
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
+NengoBones is made available under a proprietary license, the
+"ABR TECHNOLOGY LICENSE AND USE AGREEMENT" (the "ABR License"). The
+ABR License is available for download at
+`<https://www.appliedbrainresearch.com/license>`_. The ABR License is
```
drasmuss, line 12: Can we include a brief summary here (something to convey that it's free for non-commercial use)? That license file is a lot to dig through, if I saw that I would basically just say "I'm not confident parsing this long legal document, I'll use a different package".

LICENSE.rst
```diff
@@ -6,23 +6,20 @@ NengoBones license
...
+included herein by reference. The following terms and conditions, if
+any, herein are part of the ABR License.
+
+Note, if you would like to use NengoBones commercially,
+licenses can be purchased from the ABR store at
+`<https://www.appliedbrainresearch.com/store>`_. If You have any sales
```
drasmuss, line 18: You->you

tbekolay: It's deliberate and copy/pasted from the master document. I believe it is to differentiate between the colloquial and legal definition of "you".

drasmuss: We use lower case "you" in all the other places though (e.g. the previous sentence).

tbekolay:
OK, this PR's now been updated to hopefully improve the language of the ABR license summary, and to change how the license is specified in `.nengobones.yml`. I also simplified (?) some tests while doing this as in an earlier version of these changes I required the license to always be specified, but that's the case anymore, so if some of those test changes don't seem like improvements we can revert them.

nengo_bones/config.py
```diff
@@ -147,7 +147,7 @@ def fill_defaults(config):  # noqa: C901
 
     if "contributors_rst" in config:
         cfg = config["contributors_rst"]
-        cfg.setdefault("nengo_list", license_type == "nengo")
+        cfg.setdefault("nengo_list", config["license"].startswith("abr"))
```
drasmuss, line 150: I don't think we need this field any more right? We're just directly checking the license field in the contributors template.

nengo_bones/templates/LICENSE.rst.template
```diff
@@ -6,27 +6,30 @@
...
+
+{{ project_name }} is made available under a proprietary license, the
+"ABR TECHNOLOGY LICENSE AND USE AGREEMENT" (the "ABR License").
+The main ABR License file is available for download at
+`<https://www.appliedbrainresearch.com/license>`_.
+The entire contents of this ``LICENSE.rst`` file, including any
+terms and conditions herein, form part of the ABR License.
+
+Commercial Use Licenses are available to purchase for a yearly fee.
+Academic and Personal Use Licenses for {{ project_name }} are available at
```
drasmuss, line 20: Very minor/ignorable, but I'd probably just remove `for {{ project_name }}` (it's already mentioned in the title and previous paragraph).

drasmuss (approved):
Just two minor comments I can do on merge if they sound ok to you.

nengo_bones/templates/LICENSE.rst.template
tbekolay: I put it there is as I worried that people would think all ABR licensed products have a no cost option; I didn't put it in the previous sentence because it's true for all products. The rest of the license is the same for all so I wanted to be extra clear that not all products have a no cost option, but for this project, there is one.

nengo_bones/config.py
tbekolay: Ah yes, you're right! If you're fine with that change then yes let's remove this option.

PR #161, opened 2022-01-13
Automatically fill setup.py license fields

hunse:
**Motivation and context:**

In looking over https://github.com/nengo/nengo-loihi/pull/325, realized there are a few places where the license type changes things in `setup.py`, and it'd be nice to have these done automatically (unless overridden by the user). This should be particularly helpful to those setting up new projects who may forget to fill these fields otherwise. Also, the previous default for the `license` field didn't check the license type in `license_rst`, so would be wrong for Apache or MIT licenses.

Also, fixed `test_license`, which was accidentally left broken by #160.

**How has this been tested?**

Tested with https://github.com/nengo/nengo-loihi/pull/325. 

**Types of changes:**

**Checklist:**

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -182,7 +182,6 @@ setup_py:
     - "Development Status :: 4 - Beta"
     - "Framework :: Nengo"
     - "Intended Audience :: Developers"
-    - "License :: Free for non-commercial use"
     - "Operating System :: Microsoft :: Windows"
     - "Operating System :: POSIX :: Linux"
     - "Programming Language :: Python"
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -65,6 +65,8 @@ Release History
   (`#138`_)
 - Switched to new version of Codecov uploader and added checksum verification. (`#150`_)
 - Increased minimum ``black`` version to 21.12b0. (`#104`_, `#159`_)
+- License fields in ``setup.py`` are now set using ``license_rst.type``. (`#161`_)
+- Trove classifiers in ``setup.py`` are now sorted alphabetically. (`#161`_)
 
 **Fixed**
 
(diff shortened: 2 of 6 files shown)


hunse:
What was wrong with the old way of doing the Apache license's ReST? It had looked fine to me when I rendered it in HTML originally (mind you, I only took a cursory look). Now, it doesn't look quite as nice IMO: the top 3 lines of the reproduced license are no longer centred, and the top line (`Apache License`) appears in monospaced font but nothing else does, which looks a little strange.

The other changes look good to me.

tbekolay:
> What was wrong with the old way of doing the Apache license's ReST?

When I looked (both on Github and building the docs locally) the `|` character showed up at the start of each line, which isn't supposed to happen because it was supposed to be parsed as a [quoted literal block](https://docutils.sourceforge.io/docs/user/rst/quickref.html#literal-blocks). Now it has leading spaces so (at least for me building the docs locally) it is correctly parsed as an indented literal block.

> Now, it doesn't look quite as nice IMO: the top 3 lines of the reproduced license are no longer centred, and the top line (Apache License) appears in monospaced font but nothing else does, which looks a little strange.

It should look exactly like this: https://www.apache.org/licenses/LICENSE-2.0.txt

That's what it looks like to me locally. Where are you seeing that it doesn't look like that?

hunse:
I used pandoc to turn it into HTML, and this is what I got (had to add the .txt extension for Github to allow me to upload):
[license.html.txt](https://github.com/nengo/nengo-bones/files/7873378/license.html.txt)

If it looks fine in Github and when built locally, that's what we really care about I guess, so probably fine to go ahead and merge it.

I did just notice that it looks like coverage tests are failing, though.

tbekolay:
> I used pandoc to turn it into HTML, and this is what I got (had to add the .txt extension for Github to allow me to upload):
> [license.html.txt](https://github.com/nengo/nengo-bones/files/7873378/license.html.txt)

Hm interesting, how did you convert it? I did
```
pandoc -t html -f rst LICENSE.rst > LICENSE.html.txt
```
and I get this: [LICENSE.html.txt](https://github.com/nengo/nengo-bones/files/7873465/LICENSE.html.txt)

which, to be fair, is still wrong, but differently wrong. I'm guessing pandoc doesn't handle ReST very well.

I'll add a test to cover that last line and merge when codecov confirms 100% coverage :+1:

hunse:
I used that same command :shrug:

PR #165, opened 2022-07-07
Only run bones-format-notebook if docs exist

drasmuss:
**Motivation and context:**

Some projects may not have a ``docs`` folder, in which case the static script would fail when it tried to run ``bones-format-notebook docs``.

**How has this been tested?**

Used it in a project without a ``docs`` folder, static checks passed after this change.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -77,9 +77,10 @@ ci_scripts:
     pre_commands:
       - cat .ci/test-coverage.sh
       - export TEST_PRE_COMMANDS=True
-      # update nodejs version
+      - . "$HOME/.nvm/nvm.sh"
+      - nvm install 17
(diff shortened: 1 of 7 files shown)


tbekolay (commented):
This LGTM, except for the way we're getting the latest nodejs version. TravisCI has `nvm` so we should use it; I opted for the most recent pre-installed version, but we can switch to the most recent working version if you prefer @drasmuss. Will squash and merge once that's decided.

tbekolay (approved):
It's unclear what nodeJS versions are actually pre-installed, or how to access them so I opted for the same behavior as before (using v17.x) but now installed with `nvm`. Merging on green CI!

PR #167, opened 2022-08-12
Export vars in remote.sh.template

drasmuss:
This prevents shellcheck warnings in case subclassed templates don't use these variables.


tbekolay (approved):
LGTM! Merging.

PR #169, opened 2022-09-30
Allow any analytics tracking code

tbekolay:
**Motivation and context:**
In switching from UA to GA4 + Matomo, there were too many options to template out to make it worth it, so I instead switched the method of specifying analytics tags to being manual. In many ways it's easier, though more error-prone and adds bulk to `.nengobones.yml` files.

**How has this been tested?**
Used this in the NengoSphinx PR supporting this, and modified the configuration notebook and verified that it generated as expected.

**Types of changes:**

- Breaking change (fix or feature that causes existing functionality to change)

**Checklist:**


drasmuss (approved):
Added a fixup to just make the analytics code an overridable block in the `docs/conf.py` template, rather than a string in `.nengobones.yml`. My thinking was that that'd be an easier way to work with HTML, rather than as a big ol' string in a yaml file (and keeps the `.nengobones.yml` smaller).

PR #170, opened 2022-10-11
Add additional options to setup.cfg mypy section

drasmuss:
These are options that we've found useful working with mypy in downstream projects, so centralizing them here.


tbekolay (approved):
LGTM!

PR #171, opened 2022-10-20
Static check configuration updates

drasmuss:
A couple improvements to static check configuration options needed in downstream repos.

- Add blocks to pyproject.toml to make it easier for downstream projects to modify the template
- Use force_exclude instead of extend_exclude in black config so that files are ignored correctly when running through a precommit hook or IDE plugin


tbekolay (changes requested):
All looks good except the last commit message should say `extend_exclude` instead of `extend_include`.

tbekolay (approved):

PR #172, opened 2022-11-01
Remove support for TravisCI

tbekolay:
In addition, this commit:

- Fleshes out the deploy CI script to build and upload projects, which was previously done by TravisCI.
- Consolidates all bones scripts under a single `bones` command, which simplifies adding and removing scripts in the future.
- Adds a `bones check-deploy` command to check release readiness.

**Motivation and context:**
Mostly we needed the deploy script updated to migrate some downstream repos that make public releases.

**How has this been tested?**
Added tests for the new check_deploy command. Once theses tests are green, I'll also do a release candidate branch to check that the deploy to testpypi works. Maybe we should do a real release of NengoBones just for good measure?

**Types of changes:**

- Breaking change (fix or feature that causes existing functionality to change)

**Checklist:**


drasmuss (approved):
Added some minor fixups. I added the `check-manifest` and `collective.checkdocs` checks into the `check-deploy` command, so that it could be a one-stop shop for everything. And then I reverted the `JOB_SUFFIX` change, as that'll cause issues I believe (see comment below).

nengo_bones/scripts/check_deploy.py
```diff
@@ -88,6 +84,23 @@ def check(condition, msg):
...
+    output = subprocess.run(
+        ["check-manifest", root_dir],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        encoding="utf-8",
+    )
+    check(output.returncode == 0, f"Error in manifest:\n{output.stdout}")
+
+    output = subprocess.run(
+        f"python setup.py checkdocs".split(),
```
tbekolay, line 96: In the deploy script I switched to using `twine check`, which I think should do the same thing, but I don't know to what extent that works before making the actual releaes. Just trying to keep deps down :shrug:

drasmuss: Ah yeah that's even better then :+1:

nengo_bones/templates/base_script.sh.template
```diff
@@ -9,11 +9,9 @@ STATUS=0  # used to exit with non-zero status if any command fails
 # shellcheck disable=SC2034
 SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 
-if [[ -n "$SCRIPT" ]]; then
-    export JOB_NUMBER="$GITHUB_RUN_ID-$GITHUB_JOB-$SCRIPT-$GITHUB_RUN_ATTEMPT"
-else
-    export JOB_NUMBER="$GITHUB_RUN_ID-$GITHUB_JOB-$GITHUB_RUN_ATTEMPT"
-fi
+export JOB_NUMBER="$GITHUB_RUN_ID"
```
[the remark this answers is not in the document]
drasmuss: I'd originally intentionally left that out because I wanted re-running jobs to use the same job number. But looking at how we're using `JOB_NUMBER`, I don't think that's a requirement (and it might even be better where we use it for things like directory names, so we're using a clean directory rather than overwriting the contents of the previous directory). So yeah I think we might as well add in `GITHUB_RUN_ATTEMPT`.

tbekolay:
Fixups LGTM!

PR #173, opened 2022-11-07
Restrict setuptools version to <64

drasmuss:
Setuptools 64 changes how they do editable installs, which breaks a lot of code inspection tools (mypy, pylint, various IDEs). See https://mail.python.org/archives/list/typing-sig@python.org/thread/IIVBPYDZR5T5BGPAWFVYS5ZPYDXGVHQN/#OSWHT5VSRGKPSPYD7PQWR2M4OCSL5WO3 for discussion. All we can do is wait for the rest of the Python ecosystem to settle on a solution, and then for those various tools to implement that solution, and then we can remove this restriction.


tbekolay (approved):
LGTM! Merging.

PR #175, opened 2022-12-19
Update SSH conditional input syntax

drasmuss:
Github Actions changed something about their syntax so the previous way we were doing things stopped working (it would always evaluate to true).


tbekolay (approved):
The docs don't suggest any change to how paths work in coverage, but as long as this works I guess.

PR #177, opened 2023-01-03
Update black version

drasmuss:
The previous minimum version was a beta release (`black` only had beta releases at the time). This causes `pip` to install pre-releases by default. Changing to a stable (non-beta) minimum ensures that pip will only install stable releases by default.


tbekolay (approved):
LGTM!

PR #178, opened 2023-01-09
Increase minimum versions of docs packages

tbekolay:
Having outdated requirements can cause pip to spend too much time determining the right version to install. The Sphinx requirement was removed so that we can let NengoSphinxTheme handle the correct Sphinx version that it supports.

Not actually needed for KerasSpiking, but https://github.com/nengo/keras-spiking/actions/runs/3831255986/jobs/6576665494#step:5:1219 was the motivation for this.


tbekolay (commented):
Fixups LGTM!

drasmuss (approved):

PR #179, opened 2023-01-12
Switch to mambaforge for remote environments

drasmuss:
This should simplify and speed up setting up Python environments in remote builds.


tbekolay (approved):
LGTM!

PR #181, opened 2023-02-02
Mark dev version line as uncovered

drasmuss:
Without this the coverage shows up as red on release builds, which isn't the end of the world but is a bit irritating.


tbekolay (approved):
LGTM! I'll merge.

PR #183, opened 2023-02-13
Set calver version statically

tbekolay:
**Motivation and context:**

This fixes an issue in which pip would refuse to install a project using calver.

Importantly, this also changes `bones check` to allow the version to be outdated, but only for non-release versions.

**How has this been tested?**
Added a test to ensure the change to version.py doesn't fail bones check.

**Types of changes:**

**Checklist:**


nengo_bones/scripts/check_bones.py
```diff
@@ -12,6 +12,74 @@
...
+            fromfile=f"current {filename}",
+            tofile=f"new {filename}",
+        )
+    )
+
+    if (
+        "dev = 0\n" in current_lines
+        and filename == "pkg/version.py"
+        and len(diff) == 11
+        and diff[6].startswith("-version_info")
```
drasmuss, line 61: What about if instead of having this special case code for the version, we add something like a `# bones: ignore` in-line flag? That would allow for a simpler check here, and might be generally useful in other circumstances.

tbekolay:
Implemented the `# bones: ignore` option and included a changelog entry for it, ready for re-review @drasmuss

drasmuss (approved):
Fixups LGTM :+1:

PR #184, opened 2023-04-04
Avoid docformatter 1.6

drasmuss:
It introduces formatting changes that are incompatible with pep8 (see https://github.com/PyCQA/docformatter/issues/161).


drasmuss:
> LGTM, is there an open issue for the conflict somewhere on their issues list, or should we open one?

This one documents the issue https://github.com/PyCQA/docformatter/issues/161

PR #186, opened 2023-05-15
Switch to worksteal xdist distribution method

drasmuss:
This should be a generic improvement to test speed across the board. It is possible that the change to 3 workers may cause some tests suites to fail, if they have tests that had undiscovered issues with parallelization.


tbekolay (approved):
LGTM! The commit message for the last commit is incorrect, I'll amend it during the merge.

PR #187, opened 2023-06-09
Skip pypirc checks if file does not exist

vsheska:
**Motivation and context:**
We would like to run `bones check-deploy` on repos that do not have a pypirc file, such as nengo-edge-server.

Should not interact with other changes

**How has this been tested?**
I'm unsure how this should be tested.

**Types of changes:**

**Checklist:**

**Still to do:**

### DIFF
--- a/nengo_bones/tests/test_check_deploy.py
+++ b/nengo_bones/tests/test_check_deploy.py
@@ -128,11 +128,24 @@ def test_subject(tmp_path, monkeypatch):
     assert "Commit subject should be 'Release v1.2.3'" in result.output
 
 
-def test_pypirc(tmp_path, monkeypatch):
+def test_no_pypirc(tmp_path, monkeypatch):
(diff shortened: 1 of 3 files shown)


nengo_bones/tests/test_check_deploy.py
```diff
@@ -131,10 +131,8 @@ def test_subject(tmp_path, monkeypatch):
 def test_pypirc(tmp_path, monkeypatch):
     _monkeypatch(tmp_path, monkeypatch)
     result = _write_repo_and_check(tmp_path, pypirc=False)
-    assert_exit(result, 1)
```
drasmuss, line 134: We should test both cases (pypirc exists and does not exists). You can make a mock pypirc file to test the case where it exists.

drasmuss (commented):
Just the one test comment, approach looks good :+1:

PR #188, opened 2023-07-14
Use micromamba for setting up python environments

drasmuss:
This makes it easier to install packages through conda in CI setups.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,10 @@ on:
         required: false
         default: false
 
+defaults:
+  run:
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -15,6 +15,10 @@ on:
         required: false
         default: false
 
+defaults:
+  run:
+    shell: bash -el {0}
+
```
jchalatu, line 21: This change is easy enough to port over to all the other shells but it'll be good to keep an eye on the mambaforge actions repo to see if they come up with a clever way to avoid this.

PR #189, opened 2023-09-27
Adding proprietary information notice check

bjkomer:
**Motivation and context:**

This PR creates the command `bones check-notice <root_dir>` which checks all `.py` files within `<root_dir>` and its sub-directories for a proprietary information notice at the top the file. The `--add-notice` flag will insert this notice to any files that do not have it.

In addition, any files auto-generated by `nengo-bones` will include this notice if `license: proprietary` is set in the `.nengobones.yml`

**How has this been tested?**

Added `test_check_notice.py`

**Types of changes:**

**Checklist:**

**Still to do:**


nengo_bones/config.py
```diff
@@ -5,6 +5,30 @@
...
+# Patents, patents in process, and are protected by trade secret or
+# copyright law. Dissemination of this information or reproduction of
+# this material is strictly forbidden unless prior written permission is
+# obtained from Applied Brain Research. Access to the source code contained herein is
+# hereby forbidden to anyone except current Applied Brain Research employees,
+# contractors or other outside parties that have executed Confidentiality
+# and/or Non-disclosure agreements explicitly covering such access.
+#
+# The copyright notice above does not evidence any actual or intended
+# publication or disclosure  of  this source code, which includes
```
tbekolay, line 20: There's extra spaces in between words here, which are likely from some original source that tried to have everything justified even though we're in plaintext. I would remove those extra spaces and use some tool to wordwrap it to 80 characters (we usually allow more, but 80 to be safe is probably best).

drasmuss (commented):
I reworked this so that instead of a new CLI command, you set an `add_to_files: true` option in the `license_rst` section. When that is set, `bones generate` will automatically add the license text (whatever the selected license is) to the top of all python files, and `bones check` will verify that the text is there.

Also fixed an issue I ran into when working on this where `black`/`docformatter` weren't loading their config files properly when commands were being run from outside the root directory (e.g. when running tests).

nengo_bones/scripts/check_notice.py
```diff
@@ -29,59 +30,32 @@ def check_notice(root, add_notice=False, verbose=False):
...
     """
 
-    notice_lines = NOTICE.splitlines(keepends=True)
+    click.echo("Checking for license text in python files:")
+
     checked = 0
     missing = 0
     for path, _, files in os.walk(root):
         for name in files:
             if name.endswith(".py"):
```
tbekolay, line 39: Should we be doing more than just Python files? NengoLoihi has some C / C++ files. But it's true that NengoBones is really only set up to work with Python repos so probably this is fine.

tbekolay (approved):
This LGTM, one question and one potential improvement which you can do in the merge @drasmuss (or not do if you disagree, not a big thing).

nengo_bones/tests/test_config.py
```diff
@@ -10,7 +10,7 @@
 
 
 def test_find_config():
-    assert config.find_config() == pathlib.Path.cwd() / ".nengobones.yml"
+    assert config.find_config() == Path(__file__).parents[2] / ".nengobones.yml"
```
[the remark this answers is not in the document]
tbekolay: 🤯

nengo_bones/scripts/check_notice.py
drasmuss: Yeah I was thinking about that, but a lot of our C/C++ files are actually templated, so this wouldn't work for them anyway. So I thought we might as well just keep things simple here.

PR #191, opened 2023-10-06
Update versions of tools in pre-commit

lgaud:
**Motivation and context:**
isort had a problem installing in new environments due to a poetry update. The others were fairly out of date and black introduced a pre-commit mirror which is supposed to be faster.

These versions all have Python 3.8 as their minimum runtime version.

**How has this been tested?**
I tested the versions in nengo-edge-server.

**Types of changes:**

**Checklist:**

- [ ] I have read the **CONTRIBUTING.rst** document.
- [X] I have run the test suite locally and all tests passed.


drasmuss (commented):
Was talking with Trevor, and I think we should just remove `.pre-commit-config.yaml` from the templated files. That way everyone is free to set their commit hooks up however they want for their development environment. As long as commits end up passing the static checks I don't think we need to care in nengo-bones how they get there. So added a commit stripping all that stuff out.

PR #196, opened 2024-01-19
Add functionality to skip files during isort

jchalatu:
**Motivation and context:**

When running the `static` script, `isort` can run for an extremely long period of time if there are directories inside the project with large numbers of non-Python files (e.g. text files, binaries, etc.). Adding the ability to skip these directories/files with patterns alleviates this problem.

**How has this been tested?**

Added a new test and ran the entire suite locally.

**Types of changes:**

**Checklist:**


nengo_bones/templates/pyproject.toml.template
```diff
@@ -23,6 +23,12 @@ force_exclude = '''
 [tool.isort]
 profile = "black"
 src_paths = ["{{ pkg_name }}"]
+{% if isort_exclude %}
+skip_glob = [
```
drasmuss, line 27: I think we should use `extend_skip_glob` here so that it appends to the default exclusions rather than overwriting them.

While I was looking at the docs I also noticed the `skip_gitignore` option. I think we should probably enable that. That's the default behaviour for `black`, so that makes them consistent in that regard, and likely enabling that option would eliminate the need to use `isort_exclude` except in special circumstances.

jchalatu: Ah okay nice, I didn't see the `extend_skip_glob` option. 

I was initially trying to use the `skip_gitignore` flag but I couldn't get it to find the `.gitignore` file in the project directory. Although, it's possible that the regular expression that was in the `.gitignore` for my use case wasn't compatible with what `isort` expects. I agree that it would be much cleaner to use the `.gitignore` rather than customizing the exclusions (I imagine all of the relevant exclusions would be found in the `.gitignore` anyways).

jchalatu: I've pushed the fixup for adding the `skip_gitignore` flag but I'm still in the midst of seeing if I can get it to work in my use case. It works with `extend_skip_glob` at the moment, but ceases to work if I remove that from the config.

jchalatu: Ah the problem is that `skip_gitignore` traverses and checks each file that satisfies the patterns in `.gitignore` and THEN skips them. Whereas, `extend_skip_glob` just bakes those patterns into the initial traversal. See this issue for reference https://github.com/PyCQA/isort/issues/1912#issuecomment-1221669346.

drasmuss: Ah that's annoying. Hopefully they fix that at some point, but won't hurt to have the `isort_exclude` option anyway.

PR #197, opened 2024-02-06
Update actions to latest versions

drasmuss:
Github is deprecating support for Node 16 (which was previously used by most of the github actions), so we need to update to the latest versions (which use Node 20).


tbekolay (approved):
:+1: :shipit:

PR #198, opened 2024-02-26
Add option to disable caching in setup action

drasmuss:


actions/setup/action.yml
```diff
@@ -9,6 +9,10 @@ inputs:
     description: Python version (passed to actions/setup-python)
     required: false
     default: "3"
+  cache:
+    description: Enable/disable caching of pip/conda downloads
+    required: false
+    default: "true"
```
tbekolay, line 15: My gut reaction is to make this of type `boolean` and remove the quotes, but from looking at `mamba-org/setup-micromamba` it seems they don't set a type so I would assume that defaults to string.

tbekolay (approved):
LGTM!

actions/setup/action.yml
drasmuss: That's what I thought at first, but turns out you can't actually have boolean input types for composite actions (see https://github.com/actions/runner/issues/2238).

tbekolay: 😮

